PR #22, opened 2015-01-13
Switch to a more robust way of changing SELinux status

djjudas21:


jfryman:
:metal: thanks for the code

PR #25, opened 2015-01-14
REPLACE config file route

rmacian:
The original route is a symbolic link, is better to replace the value on the original file. I've found some boxes where the symbolic link does not exist and the module fails


jfryman:
Excellent. Thanks!

PR #26, opened 2015-01-16
reverting previous lint change

robrankin:
Without the quotes, Puppet 3.7 breaks.


jfryman:
How unfortunate. I'll look into this. Thanks for the fix. :wrench:

PR #28, opened 2015-01-28
fix change-selinux-status in case of selinux disabled

cristi1979:
If selinux is disabled, setenforce will fail when trying to change the status. This can be changed only with a reboot.


jfryman:
:metal: Thanks for this.

PR #29, opened 2015-01-29
Better compatibility with CentOS 7

djjudas21:
Since Facter 2.2 you can use $operatingsystemmajrelease which avoids the stupid numbering used by CentOS 7 and means you don't have to match a regex


jfryman:
Good deal. Thanks!

PR #30, opened 2015-02-04
Feature: add support for templating .te files

gavinrogers:
Hi,

I needed to be able to support this, and I think it might be useful for others. Basically, source doesn't support templates, meaning .te files can only be static. Using content, we can still use the file() function, but also use template(). I've changed the README to reflect the change. Please ping me if there's any additional garnishes that you'd like.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -48,7 +48,7 @@ Deploy a custom module:
 ```puppet
 selinux::module { 'resnet-puppet':
   ensure => 'present',
-  source => 'puppet:///modules/site_puppet/site-puppet.te',
+  content => template(modulename/module.te',
 }
 ```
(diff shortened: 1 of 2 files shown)


jfryman:
Nice. Thanks.

I like the change, but I think this should be an AND, not an OR. It will break existing users with this change.

gavinrogers:
Ok, done. I've left the documentation as showing the content option (in the interest of simplicity), but now source will work equally well, for backwards compatibility reasons and also edge cases (which I can't really imagine at this stage).

jyaworski:
Hello:

Is this still needed? If so, can you rebase?

jfryman:

PR #31, opened 2015-02-24
add spec tests, update validations, cleanup

jlambert121:
This was intended to just add spec tests, but writing spec tests highlighted some cleanup within the classes I thought needed to be done.  Sorry for the size.

Enable travis-ci support
Add spec tests
Use validators from stdlib
Move OS detection logic to params class
Ensure selinux class is included before defines
Ensure helper classes cannot be called directly


manifests/fcontext.pp
```diff
@@ -63,50 +63,42 @@
...
-    path => '/bin:/sbin:/usr/bin:/usr/sbin',
-  }
-  if $pathname == '' {
-    fail('pathname must not be empty')
-  }
 
-  if ( $equals == false ) and ( $pathname == '' ) {
-    fail('pathname must not be empty')
+  if !defined(Class['selinux']) {
+    fail('You must include the selinux base class before using any selinux defined resources')
```
jfryman: Why fail? Can the base class just not be automatically imported for them?

manifests/fcontext.pp
```diff
@@ -63,50 +63,42 @@
...
+  validate_bool($filetype, $equals)
+
+  if $equals {
+    validate_absolute_path($destination)
   }
 
-  if ( $filetype == true ) and ( $filemode == '' ) {
-    fail('file mode must not be empty')
+  if $equals and $filetype {
+    fail('Resource cannot contain both "equals" and "filetype" options')
```
jfryman: I personally am not a huge fan of letting modules arbitrarily stop processing the entire catalog if their own domain isn't covered. How about wrapping with a `warning()` and then not process the catalog?

Gemfile.lock
```diff
@@ -1,165 +0,0 @@
-GEM
```
jfryman: Why does this entire file go away?

manifests/fcontext.pp
jlambert121: That is a possibility, but I've always preferred making it explicit what modules were going to do.  Many of the resources required selinux::package which would also fail with an undefined resource, this just makes it a bit more explicit.

manifests/metadata.json
```diff
@@ -1,20 +0,0 @@
-{
```
jfryman: This is needed for Puppetforge. Any reason it's removed?

manifests/fcontext.pp
jlambert121: This is what the code did previously, just simplified it a little bit.  I'm not a fan of warnings as they are typically ignored by people - always thought validations should be fatal if you can't complete what was requrested.

manifests/port.pp
```diff
@@ -36,20 +36,17 @@
...
+  $port,
   $protocol = undef,
-  $port     = undef,
-  $policy   = 'targeted'
 ) {
-  Exec {
-    path => '/bin:/sbin:/usr/bin:/usr/sbin',
-  }
 
-  if ( $context == undef ) or ( $port == undef ) {
```
jfryman: I would prefer this stays. That way, the module controls the error message rather than Puppet throwing an error with less than useful info.

Gemfile.lock
jlambert121: To enable travis support (testing many versions of puppet and facter) you cannot have a Gemfile.lock because travis then cannot install different versions of puppet to test against.

manifests/metadata.json
jlambert121: The one in the root directory is needed by the forge, the one in the manifests directory is incorrect (and out of date looking at the version)

jfryman: Derp... yeah... I see. Sorry. :)

manifests/port.pp
jlambert121: The error that puppet throws is that "context is required" or "port is required".  This is how I've seen all of the large modules enforce required parameters.

Gemfile.lock
jfryman: :cool:

manifests/fcontext.pp
jfryman: All the same, I prefer the warning. Convergence of a system shouldn't break because of a misconfig... and admins should be investigating.

manifests/fcontext.pp
jfryman: If the base class was heavier, I'd agree. As it stands, the user really only is specifying the mode `selinux` is going to be in, so it's not too much of a stretch to provide a sane default and attempt to declare the class if it hasn't been before.

manifests/fcontext.pp
jlambert121: Should and do are different beasts. :)

I haven't seen anywhere else in this module that follows that pattern, could you give me an example of what you're thinking?

jfryman:
All in all, this is very nice and much needed. Thank you very much for adding.

A few comments we're going back and forth on... my guiding factor is to try and make modules super user-friendly which is the backing behind a few of my comments. Sane defaults and non blocking behavior are preferred in my book.

manifests/fcontext.pp
jfryman: Not so much in this module, but others I'm working on (mainly `puppet-nginx`). Also admit this might be a bit of scope creep, so maybe it's better to :football: on this one and re-address after this lands.

manifests/port.pp
jfryman: Yeah, I admit that the error messages here suck. I'm happy to merge this as-is, but I will more than likely change the behavior back in short order. So, up to you if you want to help shape it. :grinning:

jlambert121:
I updated the requirement that the selinux class is included outside the defines explicitly (just includes them now).  

Regarding wrapping things in warnings (after thinking about it for a while and talking with some other puppet guys), unless you don't want to use any stdlib validators (or puppet 4 typing) I think errors are the way to go.  If you write all of the validators yourself you can achieve a consistent validations with warnings, but leveraging any stdlib validators will fail.  Having some configurations that error (stdlib validators) and some that just warn (custom validators) will result in inconsistent feedback to the admin, but both of them will fail to apply catalogs as expected.

jlambert121:
I see you enabled travis too - nice!

jfryman:
> Regarding wrapping things in warnings (after thinking about it for a while and talking with some other puppet guys), unless you don't want to use any stdlib validators (or puppet 4 typing) I think errors are the way to go. If you write all of the validators yourself you can achieve a consistent validations with warnings, but leveraging any stdlib validators will fail. Having some configurations that error (stdlib validators) and some that just warn (custom validators) will result in inconsistent feedback to the admin, but both of them will fail to apply catalogs as expected.

Ok, then let's punt on this for now. I still may revisit this, but time is not on my side, and these changes are :+1:, so let's roll.

PR #32, opened 2015-03-18
Fix to work with strict_variables=true

robinbowes:
If selinux is disabled, the $::selinux_current_mode variable is undefined and throws an error when run with strict_variables=true

Getting the value with getvar fixes this.


jfryman:
Cool! Thanks for the code.

PR #33, opened 2015-03-25
Workaround for RH 7 variants

robrankin:
Small workaround to fix the module running on RH 7 variants, detailed in [issue 27](https://github.com/jfryman/puppet-selinux/issues/27)

Provide a new custom fact that will only execute on RH 7 variants with SELinux enabled, and use that to determine the policy in module.pp if the default selinux_config_policy fact is not either 'targeted' or 'strict'.


crashcover:
Is it possible to merge this pull request? ;-)

jfryman:
@crashcover Sure! There you go!

PR #34, opened 2015-03-27
Use "defined" instead of "getvar" to protect against undefined variables when strict_variables=yes

robinbowes:
I got into a discussion with @richardc about strict_variables and getvar.

tl;dr: use "defined" not "getvar" to protect against undefined variables.


jfryman:
:+1:

Also, +:100: debates that end in better code. :bow:

PR #35, opened 2015-04-23
Allow Scientific Linux as a valid OS

dlevene1:
Scientific Linux doesn't have a operatingsystemmajrelease so we
can't just look for v 5/6/7 etc, we need to query operatingsystemrelease

It may be better to query all of these on operatingsystemrelease
but I don't have a redhat box to query to confirm this works.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -12,21 +12,21 @@ class selinux::params {
 
   case $::osfamily {
     'RedHat': {
-      case $::operatingsystemmajrelease {
-        '7': {
+      case $::operatingsystemrelease {
+        /^7/: {
           $sx_fs_mount = '/sys/fs/selinux'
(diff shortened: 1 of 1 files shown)


belminf:
Any reason why you didn't add the period after the major release number? I suppose the use-case of RH 50 is improbable but don't see any reason not to address it in the regex.

dlevene1:
Nope - and I agree adding the period after the major release number will
future proof it.

jyaworski:
Hello:

Can you rebase and add the dot to the regexp?

dlevene1:
This can actually be dropped - the issue was an older version of facter that didn't have support for Scientific Linux operatingsystemmajrelease facter. We have since updated all our hosts and the original code now works.

PR #41, opened 2015-06-03
improve port match

kveerman:
improved port match. 

Fixes https://github.com/jfryman/puppet-selinux/issues/39


jfryman:
Nice. Thanks for the code.

PR #42, opened 2015-06-15
puppet 4 support

jhoblitt:


jhoblitt:
I just noticed that #36 also proposes to fix the lint warnings.  However, I'd claim that using control comments is a slightly better approach.

jfryman:
:bow:

jhoblitt:
:raised_hands:

PR #50, opened 2015-08-05
Added Restorecon to puppet-selinux module

meghnakale:
I used the semanage fcontext command from the puppet-selinux module (with httpd_log_t context) applied to a log folder that had two log files inside it.
After doing ls-LZ the context was reflected on the folder but it was not applied to the internal log files. 
After a research on net I found that a restorecon command is required to run for applying the context changes.
So I have created a pull request to add the restorecon command to Puppet-selinux module. Can you please review it.

### DIFF
--- a/manifests/restorecon.pp
+++ b/manifests/restorecon.pp
@@ -1,0 +1,38 @@
+# Definition: selinux::restorecon
+#
+# Description
+#  This method provides the restorecon command. After running the restorecon, the file will have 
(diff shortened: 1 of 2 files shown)


jyaworski:
Hello:

Is this still needed? If so, can you rebase?

Wouldn't it also be better to manage this with restorecond, rather than execs?

jfryman:

PR #53, opened 2015-08-28
Add Permissive to puppet-selinux module

jewnix:
Run 'semanage permissive -a' to set a context to permissive.

I use this to allow oddjob_mkhomedir_t to create home directories upon login if they do not exist.

I am not a puppet developer, so if I missed something, please review it.


jfryman:
Looks good. Thanks for the contribution!

PR #54, opened 2015-09-19
Fallback to lsbmajdistrelease, if puppet version is < 3.0

jkroepke:
Hello,

we use the new Sattelite 6.1 with Red Hat supported puppet packages.

For RHEL5, the package version is 2.7.23.

The facter operatingsystemmajrelease is named on 2.7.X, a simply fallback to lsbmajdistrelease is used, if the operatingsystemmajrelease is empty.


jfryman:
Works for me! Thanks for the code.

PR #55, opened 2015-09-19
Add support for Factor 1.6

Gilum:
The repository RHEL 5 & 6 uses Facter 1.6


jyaworski:
What versions of ruby are supported there? We look to be dropping older support in #73.

jfryman:
I am happy to merge this, but please be aware: support for older versions of non-supported Puppet/Factor is best effort.

nexecook:
Will this be merged?

PR #57, opened 2015-09-29
fcontext command escaping

Joris-van-der-Wel:
I was having some issues with `selinux::fcontext`, certain rules were re-added on every run (the `unless` property was not matching) and other rules were never added (the `unless` property was matching the wrong rule). The fcontext rules were being interpreted as a regular expression by grep.

So I added shell and regular expression escaping. Shell is simple if you use single quotes, all you have to do is replace single quotes with a double quote string that contains a single quote, for example `grep 'foo'"'"'bar'`. Regular expressions are harder, however perl regular expressions have an escape syntax using `\Qfoo\E`, all that you have to escape yourself is `\E`. That is why I am using the `-P` mode for grep.

I also changed the match itself because the `unless` property of rule `/foo` was matching `/foo/bar`

### DIFF
--- a/lib/puppet/parser/functions/selinux_escape_perl_regexp.rb
+++ b/lib/puppet/parser/functions/selinux_escape_perl_regexp.rb
@@ -1,0 +1,21 @@
+module Puppet::Parser::Functions
+    newfunction(:selinux_escape_perl_regexp, :type => :rvalue, :doc => <<-EOS
+
+Escape a string so that it can be used within a perl regular expression to match a static string.
+All special regexp characters will be escaped.
+
+For example:
+
+    $escapedRegexp = selinux_escape_perl_regexp($mystring)
+    $escapedShell = selinux_sh_escape($escapedRegexp)
+    exec { 'foo':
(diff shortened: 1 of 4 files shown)


Joris-van-der-Wel:
Rebased & I added some tests

christofdamian:
nice. I am hitting that problem too.

jyaworski:
Hello:

Could this possibly be done with http://ruby-doc.org/core-2.2.3/Regexp.html instead of gsub directly?

Joris-van-der-Wel:
I am not entirely sure what you mean, but if you are talking about Regexp.escape: I was worried that Ruby's escape might not escape everything that is recognized by grep. I avoid that issue by using the `\Q...\E` syntax of perl' regexp (`grep -P`) instead of Regexp.escape.

logic:
FWIW, I've merged this with current master over here: logic/puppet-selinux@c55799d76cf4e5cd87e3ce6db9677fcc9b585845

jfryman:

PR #59, opened 2015-09-30
Fix for selinux::module absent case failed notify

ps-jay:
Before this, selinux::module ensure => 'absent' would fail because:
- Exec["${name}-buildmod"] wouldn't be created
- Exec["${name}-checkloaded"] therefore couldn't notify it

Also:
- Add rspec testing for selinux::module
- Add custom fact selinux_config_policy to spec_helper

### DIFF
--- a/manifests/module.pp
+++ b/manifests/module.pp
@@ -58,12 +58,21 @@ define selinux::module(
     cwd          => $selinux::params::sx_mod_dir,
   }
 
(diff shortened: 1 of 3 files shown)


jfryman:
:metal:. Thanks for the code.

ps-jay:
No worries : )

If it isn't too much trouble, could you tag a 0.2.6 release?

PR #62, opened 2015-10-20
Implements SELinux type checking and ensuring.

ElvenSpellmaker:
The current module doesn't allow enforcing of the SELinux type which can make the system fail to boot unless checked.

For example someone setting the type to `mls` without proper configuration will cause the system to not be bootable. That way I can now force `enforced` and `targeted` to ensure a bootable system and ensure that I am using the correct SELinux type.

The default is set to `targeted` as this is what most people seem to use (and doesn't affect the default option of `disabled` to my knowledge) and is the default in the config on most systems.


ElvenSpellmaker:
It's true I should probably add some tests to this...

ElvenSpellmaker:
OK I am now a bit confused because the tests pass locally. =|

ElvenSpellmaker:
Note to self, don't commit the `fixtures` folder by accident.

All's well. =D

jfryman:
Great! Thanks for the code!

PR #63, opened 2015-10-27
Whitespace lint fixes

mld:
Just a couple of whitespace fixes which lets me use this module in my environment (automatic lint enforcer etc).
Thank you for a nice module.


jfryman:
Thanks for the fixes.

Issue #64, opened 2015-10-28
When using 'module' to install selinux-module the selinux-mode is set to disabled.


wouteroostervld:
When using 'module' to install selinux module mode is set to disabled. (Probably because ::selinux-class is implicitly loaded by type module.)

Expected was that just installing a module would leave the selinux-mode as-is.

See comment on: https://github.com/jfryman/puppet-selinux/commit/86e9599a7ac25a8f8bd0f8dc2d9665ef3b697bff

wouteroostervld:
I will attempt to fix this problem later this week.

thrnio:
:+1: this confused me greatly and is poor behavior IMO, not to mention undocumented.

thrnio:
This should be changed now :)

vinzent:
this should be fixed already. please re-open if you still see this problem

PR #67, opened 2015-11-04
Switch default behavior to not manage selinux

thrnio:
This changes the default behavior for the module to not modify selinux settings unless explicitly told to. This is the desired behavior as described in #64, because otherwise using one of the defined types to manage just a specific selinux rule, but not manually declaring the `selinux` class and setting the mode explicitly causes selinux to be disabled. It is confusing (and undocumented) to use a defined type in a
module called `selinux` to set an selinux rule, and have that result in selinux getting disabled.

This changes the default behavior, but it will not change the configuration of a node in the situation where the node had the class applied already. However, it will change the behavior in the situation
where the `selinux` class was not included on a node, and then was switched to being included on the node without any parameters set (included the situation of a node newly added to Puppet).


wouteroostervld:
Great!

jfryman:
Much obliged for the code! :bow:

PR #37, opened 2015-04-23
Make port exec statement unique for protocol

dlevene1:
If you have the same port allowed on both udp & tcp the puppet manifest
will fail as you have a duplicate name, this change adds the protocol
name into port definition

PR #69, opened 2015-11-11
add protocol to port exec to avoid duplicate resource

shawn-sterling:
Hello, this is pretty much the same thing as #37 as I ran into the same problem. #37 was added in April and I'm not sure the author is coming back to fix the tests. :) Also I think protocol_port reads better than port_protocol (tcp 443, vs 443 tcp) but that's just a person preference. :)

I changed the protocol to be required instead of undef, because when I don't have it:

    # semanage port -a -t http_port_t 6666
    proto option is needed for add

I get errors, so port does not appear to be an optional parameter of 'semanage port'. So I also removed the test for no protocol.

If any errors come up in the testing I'll fix em shortly; or if you have a preference for it to be another way let me know.

### DIFF
--- a/manifests/port.pp
+++ b/manifests/port.pp
@@ -38,19 +38,15 @@
 define selinux::port (
   $context,
   $port,
-  $protocol = undef,
+  $protocol,
 ) {
 
   include selinux
 
-  if $protocol {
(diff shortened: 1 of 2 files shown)


jfryman:
@shawn-sterling thanks for the contribution. Just merged #37, so I take it you don't need this anymore?

Issue #71, opened 2015-11-15
Allow disabling of selinux package management


jyaworski:
Hello:

Currently, this module always manages the selinux package. Can we add an override so we can manage it externally?

PR #72, opened 2015-11-15
Allow custom package name and management

jyaworski:
Solves #71


jfryman:
Nice! Thanks for the contribution!

Issue #70, opened 2015-11-15
Move to selmodule/selboolean for selinux::module/boolean?


jyaworski:
Hello:

It looks like puppet has support for selmodule and selboolean types.

https://docs.puppetlabs.com/references/latest/type.html#selboolean

https://docs.puppetlabs.com/references/latest/type.html#selmodule

Should we modify the existing classes for these to take advantage of the built-in types?

earsdown:
-9999 for this idea.

1. This change is not backward compatible.
2. You've just broken systems that depend on previous functionality (like selinux::module's use_makefile parameter that disappeared without warning).
3. If we wanted to use native types, we would be using one of the other selinux modules (there are many). Instead, we're using this module because we find it's better than using the native types. Which I thought was rather obvious.

If you really want to do this, consider publishing it under a different module name. If you can't revert this change and restore previous behavior, please consider putting a gigantic notice in the README warning users not to depend on this module for anything remotely important.

jyaworski:
Hello:

This was solved in #73. Native types only install the compiled .pp files for selmodule.

Closing issue.

PR #73, opened 2015-11-15
Pivot to internal types

jyaworski:
Solves #70. Work in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,24 +7,6 @@ sudo: false
 matrix:
   fast_finish: true
   include:
-  - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 2.7.0"
-  - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 3.1.0"
-  - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 3.2.0"
-  - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 3.3.0"
-  - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 3.4.0"
-  - rvm: 1.8.7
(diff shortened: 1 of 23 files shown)


jyaworski:
Adding to this, I'd like guidance on how to handle selmodule. It looks like it takes the .pp file and installs it. Does that just replace the exec, or do we refactor this?

jyaworski:
Hmm. This is also failing on Ruby 1.8.7. Ideas?

jyaworski:
Added some initial module work and did some refactoring in the process.

Note: isn't make installed with selinux? If it isn't, I can put back use_makefile.

jyaworski:
Tag @jfryman (if you didn't get the notification)

jfryman:
It looks like this is failing on more than Ruby 1.8.7. If that's all that is failing, I think it's acceptable to deprecate as Puppet doesn't support that version of ruby anymore either.

The rest though, should pass. :)

What it looks like is that the negative lookahead in your regex is causing a problem. This wasn't able to be done until 1.9.3. Can you do https://github.com/jfryman/puppet-selinux/pull/73/files#diff-5db2390a154a0e185d52ce1303d25943R46 without the lookahead?

jyaworski:
OK, this is in a reasonable state, but don't merge yet. I'm going to do more internal testing to make sure it makes sense and it flows well. Feel free to look over it in the meantime.

jyaworski:
Just as an FYI: this likely deserves whatever you consider a major version bump. I ended up dropping Ruby 1.8.7 support, as you stated it's no longer supported by upstream puppet.

jfryman:
> Just as an FYI: this likely deserves whatever you consider a major version bump. I ended up dropping Ruby 1.8.7 support, as you stated it's no longer supported by upstream puppet.

:+1:

jyaworski:
@jfryman we're good to go.

jfryman:
@jyaworski will you rebase, I'll merge this in as well.

jyaworski:
OK, rebased.

PR #78, opened 2016-01-06
Add syncversion parameter

mhjacks:
In Fedora 23, the semodule -l option doesn't return the module list anymore, and semodule --upgrade reports that it is upgraded (to use --install instead).  So this patch makes syncversion a parameter to the selinux::module type, defaulting it to true.


jfryman:
:sunglasses: Thanks for the contribution! :bow:

Issue #88, opened 2016-03-24
selinux-module failing on RHEL 7, Makefile not there


ncsutmf:
Regarding a [comment from Pull Request #73](https://github.com/jfryman/puppet-selinux/pull/73#issuecomment-156832199)

It appears RedHat decided to split selinux into the selinux-policy and selinux-policy-devel packages for RHEL 7, and put /usr/share/selinux/devel/Makefile in the devel package.  I assume this would also affect CentOS 7 and recent versions of Fedora as well.

The .te file gets created but no .mod or .pp file in /usr/share/selinux.

I didn't notice it until recently because the .mod and .pp files were already there, but I think the addition of the prefix caused it to attempt to install the module under a new name and then start giving puppet errors.

RHEL 6 seams to have the Makefile in selinux-policy and as best I can tell there is no selinux-policy-devel for RHEL 6.

Thoughts on the best way to fix this?

legooolas:
I'd suggest that either:
 - The "package_name" param should be changed to allow a list of packages, which includes selinux-policy-devel on RHEL7-based distros
 - A separate "module_package_name" param is added and used to include the package only when the selinux::module definition is used.

I'm happy to provide a patch for whichever of these is preferable (or something else :) as I'm adding the package explicitly for now and it would be nice to have the selinux module handle this on my behalf.

ncsutmf:
Currently the package is policycoreutils-python which causes selinux-policy to install as well.  I have already submitted a pull request #89 to change it to policycoreutils-devel, which requires policycoreutils-python and also causes selinux-policy-devel to install.

legooolas:
Fab -- I'll clobber this as that will fix the problem.  Also gives a handy workaround of overriding the package_name to be `policycoreutils-python-devel` for now  :)

ncsutmf:
policycoreutils-devel (no 'python' on the name)

Issue #84, opened 2016-02-16
Missing package selinux-policy-devel


gauthier-delacroix:
When upgrading to 0.3.0 from 0.2.3 on my CentOS 7.2 server, I had this error: 

    Failed to call refresh: make -f /usr/share/selinux/devel/Makefile local_mypol.pp returned 2 instead of one of [0]

The `/usr/share/selinux/devel/Makefile` directory is created by package `selinux-policy-devel` which is not installed on my server. The module seems to require it so maybe it should be added as a dependancy ?

jfryman:
Would you please open up a PR? Happy to review/merge

ncsutmf:
Just noticed I had submitted a similar issue and later submitted a fix.  See Issue #88 and Pull Request #89 

Fix in the Pull Request is to change the package name for Redhat/CentOS 7 from policycoreutils-python to policycoreutils-devel.  RPM dependencies get selinux-policy-devel and policycoreutils-python installed in this case.

Issue #82, opened 2016-02-01
Fedora 23 package name changed


logic:
In Fedora 23, after the switch to python3, `policycoreutils-python` has generally been renamed to `policycoreutils-python3`. Also, most of the utilities (including the `semanage` command) have moved to `policycoreutils-python-utils`, which is probably the better package name to target.

(Yes, they're named inconsistently. I have no good explanation, this is just what it is. :wink: )

jfryman:
Ugh. Lame.

Think you could throw together a PR?

ncsutmf:
Looks like Fedora 21 and up also require policycoreutils-devel to get the /usr/share/selinux/devel/Makefile, which has policycoreutils-python-utils as a requirement.  Working on updating my pull request #89 to take that into account and that should also resolve this issue.

PR #89, opened 2016-03-29
Switch to devel package for makefile on RHEL7 and Fedora 21+

ncsutmf:
Providing this as one possible solution to issue #88 that I submitted recently.

Switching to policycoreutils-devel for Redhat/CentOS 7 to provide /usr/share/selinux/devel/Makefile which is used by selinux::module.

policycoreutils-python is a dependency, and will be included as well when policycoreutils-devel is installed.

Same effect as using selinux-policy-devel which is the package that actually contains the makefile, but used this to stay consistent.

If still receiving errors after updating module have to delete the .te file that gets copied to /usr/share/selinux folder to get the build process to run again (set to refresh only).


ncsutmf:
Latest commit should resolve #88, #84, and probably #82 as well as officially adding Fedora 22, 23, and 24 to the supported OS list.

inequity:
Just wanted to say thanks for this patch! I am working with a weird hack-around until this gets pushed through officially. Modules seem to be effectively broken on RHEL7 without this (or a similar workaround).

jfryman:
Thanks for the contribution! :bow:

PR #92, opened 2016-04-20
Fix module installation

### LINKED ISSUE #90: selinux::module works only if module name contains local_ by default
The default prefix local_ is prefixed to module name
```
selmodule { "${prefix}${name}":
```
but this default is unintuitive. For example in the provided example
```
selinux::module{ 'apache':
  ensure => 'present',
  source => 'puppet:///modules/selinux/apache.te',
}
```
the module name inside apache.te must be
```
module local_apache 1.0;
```
so that next run finds the module from `semodule --list`.

If one provides a name
```
module apache 1.0;
```
in the file, then the module will be installed with name apache and selmodule tries to search for local_apache. As it doesn't find the module, it will reinstall the module on every run.

If the idea is to prefix module files in selmoduledir, we should change
```
selmodule { "${prefix}${name}":
  ensure       => $ensure,
  selmoduledir => $sx_mod_dir,
  syncversion  => $syncversion,
}
```
to
```
selmodule { "${name}":
  ensure       => $ensure,
  selmodulepath => "${sx_mod_dir}/${prefix}${name}",
  syncversion  => $syncversion,
}
```
in module.pp

toddnni:
The fix changes the module name that is provided for selmodule from
${prefix}${name} to $name. The module shouldn't reload selinux module
anymore on every run as now selinux module name matches with the name
provided.

Should fix #90

### DIFF
--- a/manifests/module.pp
+++ b/manifests/module.pp
@@ -67,11 +67,11 @@ define selinux::module(
     command     => "make -f ${makefile} ${prefix}${name}.pp",
   }
   ->
-  selmodule { "${prefix}${name}":
+  selmodule { $name:
     # Load the module if it has changed or was not loaded
     # Warning: change the .te version!
-    ensure       => $ensure,
-    selmoduledir => $sx_mod_dir,
-    syncversion  => $syncversion,
+    ensure        => $ensure,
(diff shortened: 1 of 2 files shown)


jfryman:
Well, let's go with it! Thanks for the code!

PR #94, opened 2016-05-03
Allow specifying selinux module content

lightoze:


jfryman:
Thanks for the contribution! :bow:

PR #98, opened 2016-05-30
Added support for running restorecon after modifying file contexts

crayfishx:
The fcontext defined type sets the permanent state, but you still have to run restorecon to actually make that change take effect.   This is a proposal that by default will always run restorecon on the pathname (unless disabled with the $restorecond = false flag set)

Happy to look at better ways of doing this, but I wanted to be able to make the changes take effect without having to take a second step in puppet.


jfryman:
Thanks for the code!

PR #99, opened 2016-06-02
Add recursion support for restorecon.

Heidistein:
This commit allows you to ser a bool and will add the '-R- flag to restorecon. Restorecon will than recurse down the dir to restore the context.
There is also an update that will change the fcontext 'unless' clause. This could match too greedy.


jfryman:
Nice. Thanks for the code!

Issue #102, opened 2016-06-17
Duplicate test?


cdenneen:
@jfryman Can I ask what this line is for exactly?

https://github.com/jfryman/puppet-selinux/blob/master/spec/classes/selinux_package_spec.rb#L32
30 seems to do similar test.. so I'm trying to figure out the context it's run in.

jfryman:
@cdenneen Nope, no reason. I think this was a merge gone wrong. Thanks!

PR #103, opened 2016-06-17
Removes duplicate package test

jfryman:
This commit removes a duplicate test that is covered not but three lines above it.

Fixes: #102


cdenneen:
Thanks... also I'm curious the spec test says for RedHat 6 and 7 that `policycoreutils-python` should be installed and 7 `policycoreutils-devel`... but the params.pp shows `policycoreutils` for 5, `policycoreutils-python` for 6 and `policycoreutils-devel` for 7.
How is this passing?

PR #109, opened 2016-08-09
Fixes Issue-105 - fcontext detection with square brackets

TJM:
This fixes the issue of detecting existing contexts with square brackets using `semanage fcontext -E`, which produces much more "parsable" output. It also somewhat simplifies the "$filemode" stuff since "a" is the default. It will also properly detect if an fcontext exists with the wrong filemode and create it with the correct filemode. However, since we are not "purging" (which would probably be super dangerous), it won't *remove* any rules that are incorrect.


maage:
I've added this commit to #117. Thanks.

Issue #105, opened 2016-07-01
fcontext detection fails if pattern contains square brackets


BSipos-RKF:
As of 0.4.0 the `fcontext` parameter pathname is passed directly to `grep` in order to detect if the path is already defined as a context. Unfortunately, this detection fails if the pathname contains square brackets used in the path expression itself. Because these grep-interpreted characters are not escaped, they will be treated as actual pattern modifiers by grep rather than as literal characters.

An example is my pathname `/var/lib/trac/projects/[^/]+(/.*)?` which should be escaped prior to using as parameter to grep as `/var/lib/trac/projects/\[^/\]+(/.*)?`.

pietervogelaar:
I have this same problem. No escaping happened, so this will fail:

    semanage fcontext -l | grep -E "^/var/www(/.*)?/cache(/.*)?.*:httpd_sys_rw_content_t:"

ashishvadhau:
I have the same problem with the detection fails if the pathname contains square brackets used in the path expression itself.
Below is my example
Mkhomedir/Selinux::Fcontext[set-non-home-user-dir_type_d]/Exec[restorecond add_user_home_dir_t_/opshome/[^/]*_type_d]: Failed to call refresh: restorecon /opshome/[^/]* returned 255 instead of one of [0]

TJM:
I am having the same issue with `/web/[^/]+/\.ssh(/.*)?`... It should probably be using "single quotes"

@ashishvadhau you and I are also having a different problem too: #107 ... restorecon is failing due to unescaped input ending up on the shell, but as I noted there, its probably not a good idea to run restorecon "like this" (wish I had a better solution)

Issue #108, opened 2016-08-09
fcontext should check for the existence of $filepath before running restorecon


TJM:
Tried a "working" config out on a new host, and got an interesting error. It tries to run restorecon against a path that doesn't exist (yet). The idea of restorecon would be to "fix" existing files, so we should check for the existence of the path before trying to run restorecon against it. Perhaps something like:

```
[ -e "${filepath}" ]
```

PR #110, opened 2016-08-09
Fixes Issue-108 - restorecon should detect path

TJM:
Here is a "one line" fix to test for the existence of e path before trying to restorecon. I also added double quotes because it seemed like a "good idea" (paths with spaces)

Let me know if you want me to rebase this against my #105 fix or keep branching from "master"

### DIFF
--- a/manifests/fcontext.pp
+++ b/manifests/fcontext.pp
@@ -132,7 +132,8 @@ define selinux::fcontext (
   if $restorecond {
     exec { "restorecond ${resource_name}":
       path        => '/bin:/sbin:/usr/bin:/usr/sbin',
-      command     => "restorecon ${restorecond_resurse_private}${restorecond_path_private}",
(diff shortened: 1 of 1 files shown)


ashishvadhau:
Hi,
Thanks for looking in to the issue.
Can you please let us know if new version of selinux puppet module is going
to out with all fixes that you did .

TJM:
This should be referring to #108

TJM:
@ashishvadhau - I would give @jfryman a little while to review/accept/etc... however, if you  are using r10k (Puppetfile) you are welcome to try my "development" version ... with the understanding that this is "works for me" code, and you use it at your own risk (but thats not much different from any puppet module, right? :)

```
mod 'selinux',
    :git => 'git@github.com:LarkIT/puppet-selinux.git',
    :ref => 'larkit'
```

maage:
I've added this commit to #117. Thanks.

PR #111, opened 2016-08-10
Fixing operatingsystem for Amazon Linux

bleiva:
We had an issue with Amazon Linux servers. This fix is for avoiding the legend 'Operating system is not supported'

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -37,6 +37,17 @@ class selinux::params {
             }
           }
         }
+       'Amazon': {
+        $sx_fs_mount = '/selinux'
+          case $os_maj_release {
+            '4': {
+              $package_name = 'policycoreutils-python'
+            }
+            default: {
+              fail("${::operatingsystem}-${::os_maj_release} is not supported")
+            }
+          }
+        }
         default: {
           case $os_maj_release {
             '7': {


seefood:
I was just about to patch this in! Can you please approve the PR?

bastelfreak:
@bleiva can you rebase against latest master? This should make travis green

jyaworski:
Merging since this is so trivial. Rebasing is best practices, but people only have so much time.

seefood:
Rebasing is tricky business sine it kills many fine features of git's merge abilities. I'm a little scared of it, and all the books and manpages warn against mixing rebase and merge in the same repo, which sounds like very sensible advice to me. merge is the default in git, and in most github projects I have bumped into to date, so I'll stick to that :)

bastelfreak:
Hi @seefood, sounds interesting, I have never heard of that also never experienced issues. Do you have a link to that information?

JimPanic:
There's no problem mixing merge and rebase in the same *repository*. But you shouldn't mix merges and rebases in the same PR/branch divergence.

Rebasing a PR onto master is still best practices since it neither scrambles the history, nor any of the "fine features" take away the possibility to reason about your PR.

Not saying merging is bad or the features wouldn't be fine in most cases; in case of revitalizing PRs however, rebasing is a very safe strategy.

seefood:
@bastelfreak either you were lucky or you are aware of the possible pitfalls of rewriting commit history and manage to steer carefully around them. when rebase rewrites the history, it gives commits new hashes, so diamond merges don't have history to work with and such. So if not careful, you end up patching the same change twice and other headaches that would not happen if the history was not lost.

seefood:
Well, I had a few cases where I could not wait for a repo owner to accept a PR and pulled it into my branch, knowing it would be merged by the time I send in my PR. with rebase it would have meant the same changes would be merged with different hashes, which may mean a painful manual merge. Maybe there are better examples, I'm just glad I have not had to deal with them.

PR #114, opened 2016-09-01
Support for Amazon Linux

seefood:


jyaworski:
@seefood could you see if this is still needed? The latest master contains related code now.

seefood:
I won't have time to test it before next week, I think, but it looks correct in the source. I don't see Amazon in the metadata.json, but that's not going to matter to my system :-)

PR #117, opened 2016-09-04
Cleanups and dangling issues

maage:
This patchset contains several small cleanups. It also contains fixes for #105 and #108 from TJM. Those issues do have some issue with travis and have not gone forward. As both issues are something I'd like to have and also have further changes in shellquote part, I've added them here.

### DIFF
--- a/manifests/fcontext.pp
+++ b/manifests/fcontext.pp
@@ -70,7 +70,7 @@ define selinux::fcontext (
   $destination         = undef,
   $context             = undef,
   $filetype            = false,
-  $filemode            = undef,
+  $filemode            = 'a',
   $equals              = false,
   $restorecond         = true,
   $restorecond_path    = undef,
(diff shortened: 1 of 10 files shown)


spec/defines/selinux_fcontext_spec.rb
```diff
@@ -18,41 +18,51 @@
     it { expect { is_expected.to compile }.to raise_error(%r{file mode must be one of: a,f,d,c,b,s,l,p - see "man semanage-fcontext"}) }
   end
 
+  context 'invalid multiple filetype' do
+    let(:params) { { pathname: '/tmp/file1', filetype: true, filemode: 'afdcbslp', context: 'user_home_dir_t' } }
```
bastelfreak: could you please reformat this like:
```ruby
let(:params) do
  {
    pathname: '/tmp/file1',
    filetype: true,
    filemode: 'afdcbslp',
    context: 'user_home_dir_t'
  }
end
```

maage:
ok

manifests/fcontext.pp
```diff
@@ -70,7 +70,7 @@
   $destination         = undef,
   $context             = undef,
   $filetype            = false,
-  $filemode            = undef,
+  $filemode            = 'a',
```
bastelfreak: changing default behavior requires us to do a major release. I think this just changes the default value, but not the behavior because of the if constraint in line 107. So a normal minor release should be sufficient?

bastelfreak:
Hi @maage, thanks for the PR! Looks very good to me, I just commented two little things.

maage:
Is there rubocop setting with I can catch those wrongly formatted hashes?

About filemode, previously if you defined resource without (undef) you got no change at all if you had previous setting where filemode was something else than 'a'/undef, because grep was positive from the wrong filemode. This definitely changes what happens. But I think previous setting was buggy. Same with values where shellquote is needed. With carefully constructed fcontext you could get "working" config but at least grep would give you wrong value and maybe you'd end up running semanage -a over and over again.

I noticed slight bug when filetype was false and filemode was bogus and I've added patch to test/fix it. This will also break things.

Main issue might be #105 change where semanage looks only locally made changes and not global changes. If previously you fixed wrong upstream module with local fcontext, and then upstream got fixed and you did new install, and now module does not add resources any more. But now module adds resources regardless.

Compared to what I have in my queue, these changes are minor, but definitely there can be config that can break stuff.

bastelfreak:
rubocop: It should be able to fix it if there is at least one new line in the hash, but I am not sure. 

----

If I understand you their are certain circumstances where we changed the default behavior. So we should do a major release.

manifests/fcontext.pp
TJM: The default value is 'a' anyhow, it just makes the matching code easier

jyaworski:
@bastelfreak let's merge this, then do a major bump after the dust settles from the other outstanding PRs.

PR #128, opened 2016-11-04
Fixes Issue-104 - No Hiera support

ryayon:
Hiera hash are managed only if they exist 
+ hiera lookups outside of the variables part of the manifest.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -36,7 +36,7 @@ class selinux (
   $permissive     = undef,
   $port           = undef,
   ### END Hiera Lookups ###
-  
+
 ) inherits selinux::params {
 
   $mode_real = $mode ? {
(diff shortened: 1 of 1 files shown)


ryayon:
Multiples commit are due to the fact I don't have ssh access to GitHub from my office => I commit from GitHub WebUI directly.

manifests/init.pp
```diff
@@ -47,6 +40,13 @@
     default => 'undef',
   }
 
+  ### START Hiera Lookups ###
+  $selinux_booleans  = hiera('selinux_booleans', undef)
+  $selinux_modules   = hiera('selinux_modules', undef)
+  $selinux_fcontexts = hiera('selinux_fcontexts', undef)
+  $selinux_ports     = hiera('selinux_ports', undef)
```
jyaworski: Should these all be prefixed with `selinux::`?

PR #140, opened 2016-12-14
Default module prefix now ''

traylenator:
With RHEL7.3 the default module prefix causes problems
in particular the following happens

with

```puppet
selinux::module { 'openshift-cvmfs':
  ensure => present,
  source => 'puppet:///modules/hg_paas/openshift-cvmfs.te',
}

```

```
Compiling targeted local_openshift-cvmfs module
/usr/bin/checkmodule:  loading policy configuration from
tmp/local_openshift-cvmfs.tmp
/usr/bin/checkmodule:  Module name openshift-cvmfs is different than the
output base filename local_openshift-cvmfs
make: *** [tmp/local_openshift-cvmfs.mod] Error 1
```

As far as I can ascertain the prefix only influences the file name
and actual module name comes from what's inside the policy file.

Anyone testing this may well runinto another selinux bug that
is not a problem of this module.

* [CERN ticket reference](https://its.cern.ch/jira/browse/AI-4745)
* [Change in checkmodule](https://github.com/SELinuxProject/selinux/commit/c6acfae4bc22586ad1dc259b0aad57fa6c5b43ef#diff-9a0edd4c660062ff41e43d787595141d)
* [puppet problem](https://bugzilla.redhat.com/show_bug.cgi?id=1391886)
* [selinux bug](https://bugzilla.redhat.com/show_bug.cgi?id=1395222)

### DIFF
--- a/manifests/module.pp
+++ b/manifests/module.pp
@@ -13,7 +13,7 @@
 #   - $sx_mod_dir (absolute_path) - sets the operating state for SELinux.
 #   - $source: the source file (either a puppet URI or local file) of the SELinux .te module
 #   - $makefile: the makefile file path
-#   - $prefix: the prefix to add to the loaded module. Defaults to 'local_'.
+#   - $prefix: the prefix to add to the loaded module. Defaults to ''.
 #
 # Actions:
 #  Compiles a module using make and installs it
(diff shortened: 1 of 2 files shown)


traylenator:
There is another option to remove the prefix parameter. I think it should actually never be set.

vinzent:
@traylenator either way defaulting to unset or removing it at all is a breaking change for existing users. 

My vote to remove it completly.

bastelfreak:
Thanks @traylenator! I think keeping it for now is the better approach.

PR #144, opened 2016-12-23
Set puppet minimum version_requirement to 3.8.7

juniorsysadmin:
Also bump minimum version for stdlib (for Puppet 4) and fix invalid version_range key


juniorsysadmin:
https://github.com/voxpupuli/metadata-json-lint/issues/47

PR #145, opened 2016-12-23
Add acceptance tests

vinzent:
This PR adds a simple acceptance test scenario.

It will:

* enable selinux
* set it to enforcing
* load a custom policy with a custom domain, boolean and file and port
contexts
* enable a custom boolean
* label a port with a custom context
* set a custom context on a file and 
* it will set the custom domain into permissive mode

Run `PUPPET_INSTALL_TYPE=agent bundle exec rake beaker:centos-72-x64` to
test with CentOS 7.

Run `PUPPET_INSTALL_TYPE=agent bundle exec rake beaker:centos-66-x64` to
test with CentOS 6.

There is a known problem with CentOS 7 and the selmodule type. See 
PUP-6908, PUP-5649

### DIFF
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -1,0 +1,80 @@
+require 'spec_helper_acceptance'
+
+describe 'selinux class' do
+  let(:pp) do
+    <<-EOS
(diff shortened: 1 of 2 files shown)


alexjfisher:
Looks great.  Thanks!

PR #151, opened 2016-12-24
Create /.autorelabel when switching from disabled

### LINKED ISSUE #149: Ensure a complete relabeling when switching from disabled to permissive or enforcing

- Module version: 0.5.1-rc0

* pre-requirements:  systems has selinux disabled but selinux packages installed

```
class { 'selinux':
  mode => 'permissive',
  type   => 'targeted',
}
```

## What are you seeing

selinux configured for enforcing, but no autorelabel trigger file is created

When SELinux is disabled and desired state is permissive or enforcing the module must create the `/.autorelabel` trigger file 

If a system has selinux disabled and wants to switch to permissive or enforcing mode a complete relabeling of the filesystem is necessary as within disabled mode no filesystem attributes regarding contexts are written, thus leaving it in a inconsitent state.

vinzent:
A complete relabeling is required when switching from disabeld to
permissive or enforcing mode, because selinux file attributes
are not written in disabled mode this leaving the files regarding
selinux in an incosistent state.

This resolves #149

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -59,6 +59,18 @@ class selinux::config (
       }
     }
 
+    # a complete relabeling is required when switching from disabled to
+    # permissive or enforcing. Ensure the autorelabel trigger file is created.
+    if $mode in ['enforcing','permissive'] and
+      !$::selinux_enabled {
+      file { '/.autorelabel':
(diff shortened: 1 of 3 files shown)


bastelfreak:
thanks!

PR #153, opened 2016-12-24
(WIP) Add disabled to enforcing acceptance test

vinzent:
Simulate the situation when the systems is in disabled state
and puppet tries to configure enforcing mode which requires
a reboot to get selinux enabled.

### DIFF
--- a/spec/acceptance/while_disabled_spec.rb
+++ b/spec/acceptance/while_disabled_spec.rb
@@ -1,0 +1,93 @@
+require 'spec_helper_acceptance'
+
+describe 'while disabled' do
+  before(:all) do
+    apply_manifest('class { "selinux": mode => "disabled" }', catch_failures: true)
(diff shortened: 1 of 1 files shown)


vinzent:
On EL7 `selinux::boolean` resources fail while selinux is disabled. 

```
Error: /Stage[main]/Main/Selinux::Boolean[puppet_selinux_test_policy_bool]/Selboolean[puppet_selinux_test_policy_bool]: Could not evaluate: Execution of '/usr/sbin/getsebool puppet_selinux_test_policy_bool' returned 1: /usr/sbin/getsebool:  SELinux is disabled
```

vinzent:
On EL6 `selinux::boolean` and `selinux::permissive` fail

vinzent:
opening another PR as the scenario is not doable right now because selinux::boolean, selinux::permissive and selinux::port fail while selinux is disabled.

PR #156, opened 2016-12-26
Remove mentions of Ruby requirements in README

juniorsysadmin:
Ruby 1.9.3 is EOL, regardless of where it may be bundled.
Avoid mentions of specific Ruby versions

Also remove mentions of specific Facter versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,7 @@ This class manages SELinux on RHEL based systems.
 
 ## Requirements
 
-* Puppet-3.x or later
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
We don't have a supported Ruby matrix for all of our modules - but we test against a selection. Explicit recommendations for a requirement should be based on Puppet versions. Listing Ruby versions as a requirement can easily cause a misconception of long-term support for certain Ruby versions, when there doesn't need to be one.

PR #158, opened 2016-12-28
Default to undef for syncversion parameter in selinux::module 

vinzent:
Newer SELinux tools don't show the modules version number when
calling `semodule -l`. Unfortunatly there is a bug in puppet's
selmodule type which leads to re-applying the resource every
time even when syncversion is false. Only undef works.

This change defaults the syncversion parameter to undef instead
of true.

This fixes the repeated change:

```
Selmodule[puppet_selinux_test_policy]/syncversion: syncversion changed 'false' to 'true'
```

See also: PUP-6908, PUP-5649

Succesfully ran centos-72-x64 and centos-66-x64 acceptance tests.

### DIFF
--- a/manifests/module.pp
+++ b/manifests/module.pp
@@ -1,32 +1,27 @@
-# Definition: selinux::module
+# Defined type: selinux::module
 #
-# Description
-#  This class will either install or uninstall a SELinux module from a running system.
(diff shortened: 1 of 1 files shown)


bastelfreak:
thanks!

PR #159, opened 2016-12-28
Fix usage of non-existent $::selinux_enabled fact

vinzent:
The fact is in fact called $::selinux not $::selinux_enabled.
    
It was not caught by the spec test because it was specified in
default_module_facts.yaml.
    
 And there is only a WIP accetpance test for disabled to enforcing
switch which will check this too in future.

The existing acceptance tests for centos-72-x64 and 
centos-66-x64 discovered no problems with this PR.


bastelfreak:
thanks!

Issue #147, opened 2016-12-23
Ordering of resources


vinzent:
## Affected Puppet, Ruby, OS and module versions/distributions

- Module version: 0.5.1-rc0

## How to reproduce (e.g Puppet code you use)

```
          selinux::module { 'puppet_selinux_test_policy':
            content => "policy_module(puppet_selinux_test_policy, 1.0.0)\ngen_tunable(puppet_selinux_test_policy_bool, false)\ntype puppet_selinux_test_policy_t;\ntype puppet_selinux_test_policy_exec_t;\ninit_daemon_domain(puppet_selinux_test_policy_t, puppet_selinux_test_policy_exec_t)\n",
            prefix => '',
            syncversion => undef,
          }
  
          file { '/tmp/test_selinux_fcontext':
            content => 'TEST',
            seltype => 'puppet_selinux_test_policy_exec_t',
          }

          selinux::boolean { 'puppet_selinux_test_policy_bool': }
```

## What are you seeing

puppet tries to manage the boolean and only afterwards it loads the module. Because the module defines the boolean it fails to manage it.

## What behaviour did you expect instead

Ordering of resources should be:

![workflow](https://cloud.githubusercontent.com/assets/296297/21741963/1e7f4520-d4e5-11e6-8d74-7a8d21ec37c8.png)

(uninstall is not a feature of the module)

## Output log

```
Notice: /Stage[main]/Main/File[/tmp/test_selinux_fcontext]/ensure: defined content as '{md5}033bd94b1168d7e4f0d644c3c95e35bf'
Warning: Failed to set SELinux context unconfined_u:object_r:puppet_selinux_test_policy_exec_t:s0 on /tmp/test_selinux_fcontext
Error: /Stage[main]/Main/Selinux::Boolean[puppet_selinux_test_policy_bool]/Selboolean[puppet_selinux_test_policy_bool]: Could not evaluate: Execution of '/usr/sbin/getsebool puppet_selinux_test_policy_bool' returned 255: Error getting active value for puppet_selinux_test_policy_bool
Notice: /Stage[main]/Main/Selinux::Module[puppet_selinux_test_policy]/File[/usr/share/selinux/puppet_selinux_test_policy.te]/ensure: created
[...]

PR #161, opened 2016-12-29
Don't add selboolean resources while selinux is disabled

vinzent:
The selboolean type calls getsebool and setsebool to manage
the boolean. They only work if selinux is enabled either in
permissive or enforcing mode.

This change checks if selinux is enabled and only managages
the selboolean resource if selinux is enabled.

### DIFF
--- a/manifests/boolean.pp
+++ b/manifests/boolean.pp
@@ -1,30 +1,21 @@
-# Definition: selinux::boolean
+# Defined type: selinux::boolean
 #
-# Description
-#  This class will set the state of an SELinux boolean.
-#  All pending values are written to the policy file on disk, so they will be persistant across reboots.
-#  Ensure that the manifest notifies a related service as a restart for that service may be required.
+# This class will set the state of an SELinux boolean.
+# All pending values are written to the policy file on disk, so they will be persistant across reboots.
+# Ensure that the manifest notifies a related service as a restart for that service may be required.
 #
-# Class created by GreenOgre<aggibson@cogeco.ca>
-#  Adds to puppet-selinux by jfryman
-#   https://github.com/jfryman/puppet-selinux
+# @example activate boolean
+#   selinux::boolean{ 'named_write_master_zones':
+#      ensure     => 'on',
+#   }
 #
-# Parameters:
-#   - $ensure: (on|off) - Sets the current state of a particular SELinux boolean
-#   - $persistent: (true|false) - Should a particular SELinux boolean persist across reboots
+# @example disable boolean
+#   selinux::boolean{ 'named_write_master_zones':
+#      ensure     => 'off',
+#   }
 #
-# Actions:
-#  Wraps selboolean to set states
-#
-# Requires:
-#  - SELinux
-#
-# Sample Usage:
-#
-#  selinux::boolean{ 'named_write_master_zones':
-#     ensure     => "on",
-#     persistent => true,
-#  }
+# @param ensure Sets the current state of a particular SELinux boolean. Valid values: on, off
+# @param persistent Should a particular SELinux boolean persist across reboots
 #
 define selinux::boolean (
   $ensure     = 'on',
@@ -48,8 +39,12 @@ define selinux::boolean (
     default                    => undef,
   }
 
-  selboolean { $name:
-    value      => $value,
-    persistent => $persistent,
+  # seboolean calls getsebool and setsebool 
+  # they only work when current mode is not disabled.
+  if $::selinux {
+    selboolean { $name:
+      value      => $value,
+      persistent => $persistent,
+    }
   }
 }
--- a/spec/defines/selinux_boolean_spec.rb
+++ b/spec/defines/selinux_boolean_spec.rb
@@ -5,7 +5,12 @@ describe 'selinux::boolean' do
   on_supported_os.each do |os, facts|
     context "on #{os}" do
       let(:facts) do
-        facts
+        facts.merge(
+          selinux: true,
+          selinux_config_mode: 'enforcing',
+          selinux_config_policy: 'targeted',
+          selinux_current_mode: 'enforcing'
+        )
       end
 
       ['on', true, 'present'].each do |value|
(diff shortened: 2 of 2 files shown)


vinzent:
Haven't run acceptance tests yet because of poor internet connection.

rnelson0:
While this does prevent errors (and makes sure that the include selinux counts for something), it can be misleading that someone would add the selboolean but it ends up not being part of the desired state. Should there be an else case that notifies or fails catalog compilation when selinux is not enabled at all?

vinzent:
@rnelson0 given i'm using roles/profiles and different profiles add `selinux::boolean` and somebody overrides `$selinux::mode` at the role level to `disabled` I don't want a notify (resource change) every puppet run. 

maybe check for `$::selinux::mode == undef` and `$::selinux_current_mode == 'disabled'`for a warning?

vinzent:
Given I manage a boolean and something outside of puppet manages the SELinux mode (disabled, permissive, enforcing) I wan't no compile or runtime error and no forever changing resources (like notify).

inbetween I discovered that `semanage boolean --noreload --modify --on $boolean` works while disabled.

rnelson0:
@vinzent what's the current behavior when a selboolean is set and selinux is disabled entirely? Does the catalog compile but the manifest fails to compile cleanly, or some other result?

vinzent:
test manifest:
```
notify { "selinux=${::selinux}": }
selboolean { 'zabbix_can_network': persistent => true }
```

centos--66-x64 while selinux disabled:
```
[root@centos-66-x64 ~]# puppet apply /tmp/test.pp
Notice: Compiled catalog for centos-66-x64.local in environment production in 0.07 seconds
Notice: selinux=false
Notice: /Stage[main]/Main/Notify[selinux=false]/message: defined 'message' as 'selinux=false'
Notice: Applied catalog in 0.01 seconds
```

```
root@centos-66-x64 ~]# grep -r zabbix_can_network /etc/selinux/
Binary file /etc/selinux/targeted/policy/policy.24 matches
Binary file /etc/selinux/targeted/modules/active/policy.kern matches
```

vs. `semanage boolean --noreload --modify --on zabbix_can_network`:

``` 
[root@centos-66-x64 ~]# grep -r zabbix_can_network /etc/selinux/
Binary file /etc/selinux/targeted/policy/policy.24 matches
Binary file /etc/selinux/targeted/modules/active/policy.kern matches
/etc/selinux/targeted/modules/active/booleans.local:zabbix_can_network=1
```

Don't know where i've seen the selboolean failure. but at least it just ignores the selboolean resource when selinux is disabled. so after reboot to enforcing no boolean is set and probably services fail.

rnelson0:
Since it works fine and you can't reproduce your error, does this need changed at all then?

vinzent:
Interestingly I get it while using puppet-selinux disabled:

```
[root@centos-66-x64 ~]# /opt/puppetlabs/bin/puppet apply -e "selinux::boolean { 'zabbix_can_network': }"
<... stdlib deprecations ...>
Notice: Compiled catalog for centos-66-x64.local in environment production in 0.37 seconds
Error: /Stage[main]/Main/Selinux::Boolean[zabbix_can_network]/Selboolean[zabbix_can_network]: Could not evaluate: Execution of '/usr/sbin/getsebool zabbix_can_network' returned 1: /usr/sbin/getsebool:  SELinux is disabled
``` 

and I think every semanage exec resource also produces errors https://paste.fedoraproject.org/514921/75582148/

And I'm going mad, the same example as yesterday fails now too:

```
[root@centos-66-x64 ~]# /opt/puppetlabs/bin/puppet apply /tmp/test-selboolean.pp
Notice: Compiled catalog for centos-66-x64.local in environment production in 0.07 seconds
Notice: selinux=false
Notice: /Stage[main]/Main/Notify[selinux=false]/message: defined 'message' as 'selinux=false'
Error: /Stage[main]/Main/Selboolean[zabbix_can_network]: Could not evaluate: Execution of '/usr/sbin/getsebool zabbix_can_network' returned 1: /usr/sbin/getsebool:  SELinux is disabled
Notice: Applied catalog in 0.03 seconds
```

vinzent:
of course, yesterday I didn't specify the value parameter. `value => on`! without value it doesnt manage the boolean ... thus no error...

vinzent:
``` 
expectations while selinux is disabled:
  File "/etc/selinux/config"
    content
      should match /^SELINUX=disabled$/
  Command "getenforce"
    stdout
      should match /^Disabled$/
  some semanage --list operations fail
    Command "semanage boolean --list"
      exit_status
        should not eq 0
    Command "semanage permissive --list"
      exit_status
        should not eq 0
      stderr
        should match "/usr/sbin/semanage: No default policy found."
  some semanage --list operations work
    Command "semanage fcontext --list"
      exit_status
        should eq 0
    Command "semanage port --list"
      exit_status
        should eq 0
  getsebool and setsebool fail
    Command "getsebool"
      exit_status
        should not eq 0
      stderr
        should match "getsebool:  SELinux is disabled"
    Command "setsebool -P zabbix_can_network on"
      exit_status
        should not eq 0
      stderr
        should match "setsebool:  SELinux is disabled"
  facter facts
    Command "facter selinux"
      stdout
        should match "false"
    Command "facter os.selinux.enabled"
      stdout
        should match "false"
    Command "facter selinux_current_mode"
      stdout
        should match "\n"
    Command "facter os.selinux.current_mode"
      stdout
        should match "\n"
    Command "facter selinux_config_mode"
      stdout
        should match "\n"
    Command "facter os.selinux.config_mode"
      stdout
        should match "\n"
    Command "facter selinux_config_policy"
      stdout
        should match "\n"
    Command "facter os.selinux.config_policy"
      stdout
        should match "\n"
```

vinzent:
diff for the same test for centos-72-x64:

```
rspec ./spec/acceptance/expectations_while_disabled_spec.rb:23 # expectations while selinux is disabled: some semanage --list operations fail Command "semanage permissive --list" exit_status should not eq 0
rspec ./spec/acceptance/expectations_while_disabled_spec.rb:24 # expectations while selinux is disabled: some semanage --list operations fail Command "semanage permissive --list" stderr should match "/usr/sbin/semanage: No default policy found."
```
(`semanage permissive --list` works on EL7)

vinzent:
related issues: 

https://github.com/voxpupuli/puppet-selinux/issues/147
https://github.com/voxpupuli/puppet-selinux/issues/147

closing this now.

PR #172, opened 2017-01-13
Update inline doc to puppet-strings

vinzent:
* puppet strings requires redcarpet gem to convert markdown to html
* Github Pages require the docs to be in the docs/ folder
* the docs can be generated with: `bundle exec rake strings:generate`


rnelson0:
What does `redcarpet` do, and how was `docs/` generated? It seems like we need a rake task to automate this, and/or a blog post to go with it.

vinzent:
@rnelson0 redcarpet converts markdown to html , there is a strings rake target already: `bundle exec rake strings:generate`

Issue #164, opened 2017-01-07
Silently doesn't remove port context


vinzent:
## Affected Puppet, Ruby, OS and module versions/distributions

- Module version: v0.7.1

## How to reproduce (e.g Puppet code you use)

```
selinux::port { 'puppet_selinux_test_policy_port_t/tcp':
  context => 'puppet_selinux_test_policy_port_t',
  port => '55555',
  protocol => 'tcp',
  argument => '-d',
}
```

## What are you seeing

```
[root@centos-72-x64 ~]# puppet apply /tmp/test.pp 
...
Notice: Compiled catalog for centos-72-x64.local in environment production in 0.40 seconds
Notice: Applied catalog in 1.03 seconds
``` 

port context is still defined:

```
[root@centos-72-x64 ~]# semanage port --locallist -l
SELinux Port Type              Proto    Port Number

puppet_selinux_test_policy_port_t tcp      55555
```

## What behaviour did you expect instead

port context should be removed

## Any additional information you'd like to impart

* I don't like the `$arguments` param. it should really be something like ensure => (present|absent)
* related to https://github.com/voxpupuli/puppet-selinux/issues/147

oranenj:
Is this still an issue? Wasn't this fixed by the port provider? :)

vinzent:
fixed wiht #174

Issue #119, opened 2016-09-04
This module accepts invalid config for port


maage:
### Affected Puppet, Ruby, OS and module versions/distributions
- Puppet: Puppet v3.8.7
- Ruby: ruby 2.0.0p598 (2014-11-13) [x86_64-linux]
- Distribution: CentOS 7.2
- Module version: 0.4.1
### How to reproduce (e.g Puppet code you use)

```
ensure_resource('selinux::port', { 'foo', { context => 'syslogd_port_t', protocol => 'tcp6', port => 8514 }}
ensure_resource('selinux::port', { 'foo', { context => 'syslogd_port_t', port => 8514 }}
```
### What are you seeing

Errors from semanage command:

```
ValueError: Protocol udp or tcp is required
proto option is needed for add
```
### What behaviour did you expect instead

I'd expect manifest to catch these.
### Any additional information you'd like to impart

rpm -qf /usr/sbin/semanage
policycoreutils-python-2.2.5-20.el7.x86_64

semanage port interface is the same in  RedHat 6.8 policycoreutils-python-2.0.83-29.el6.x86_64.

```
/usr/sbin/semanage: Protocol udp or tcp is required
```

maage:
I have branch with fixes:
https://github.com/cybercom-finland/puppet-selinux/tree/port-limit-119 
But I would like to have #117 fixed before.

Issue #93, opened 2016-05-02
Unable to modify port via port.pp


pboguszewski:
I was in need to add a port to an existing SElinux port group but semange -a fails with the error: "ValueError: Port xxx/xxx already defined". To work around this I copied the port.pp to mport.pp and simply changed the '-a' for the add options to '-m' for modify.  It works as planned.  Modified code below:

define selinux::mport (
  $context,
  $port,
  $protocol = undef,
) {

  include selinux

  if $protocol {
    validate_re($protocol, ['^tcp6?$', '^udp6?$'])
    $protocol_switch="-p ${protocol} "
    $port_exec_command = "add_${context}_${port}_${protocol}"
  } else {
    $protocol_switch=''
    $port_exec_command = "add_${context}_${port}"
  }

  exec { $port_exec_command:
    command => "semanage port -m -t ${context} ${protocol_switch}${port}",
    unless  => "semanage port -l|grep \"^${context}._${protocol}._${port}\"|grep -w ${port}",
    path    => '/bin:/sbin:/usr/bin:/usr/sbin',
    require => Class['selinux::package']
  }
}

pboguszewski:
The other way to do this would be to add an 'action' parameter and default it to 'a' for add but allow for the 'm' for modify:

define selinux::port (
  $context,
  $port,
  $protocol = undef,
  $action = 'a',
)

Then in the exec statement use the action parameter:

  exec { $port_exec_command:
    command => "semanage port -${action} -t ${context} ${protocol_switch}${port}",
    unless  => "semanage port -l|grep \"^${context}.*${protocol}.*${port}\"|grep -w ${port}",
    path    => '/bin:/sbin:/usr/bin:/usr/sbin',
    require => Class['selinux::package']
  }

pboguszewski:
Or, better yet for easier readability:

`# Definition: selinux::fcontext
#
# Description
#  This method will manage a local network port context setting, and will
#  persist it across reboots.
#  It will perform a check to ensure the network context is not already set.
#  Anyplace you wish to use this method you must ensure that the selinux class is required
#  first. Otherwise you run the risk of attempting to execute the semanage and that program
#  will not yet be installed.
#
# Class create by Matt Willsher<matt@monki.org.uk>
# Based on selinux::fcontext by Erik M Jacobs<erikmjacobs@gmail.com>
#  Adds to puppet-selinux by jfryman
#   https://github.com/jfryman/puppet-selinux
[...]

Issue #38, opened 2015-05-20
Adding a port gets an error the first time


TJM:
I am seeing the following error:

```
err /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns change from notrun to 0 failed: semanage port -a -t ssh_port_t -p tcp 1022 returned 1 instead of one of [0]
err Puppet  semanage port -a -t ssh_port_t -p tcp 1022 returned 1 instead of one of [0]
notice  /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns ValueError: Could not commit semanage transaction
notice  /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns libsemanage.semanage_reload_policy: load_policy returned error code 2.
notice  /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns /sbin/load_policy: Can't load policy: No such file or directory
notice  /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns SELinux: Could not open policy file <= /etc/selinux/targeted/policy/policy.29: No such file or directory
notice  /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns SELinux: Could not downgrade policy file /etc/selinux/targeted/policy/policy.29, searching for an older version.
notice  /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns libsemanage.semanage_reload_policy: load_policy returned error code 2.
notice  /Stage[main]/Lark::Profile::Ssh/Selinux::Port[allow-ssh-port-selinux]/Exec[add_ssh_port_t_1022]/returns /sbin/load_policy: Can't load policy: No such file or directory
[...]

PR #174, opened 2017-01-14
Add a selinux_port type and provider

oranenj:
The selinux module currently handles port resources (among others) via exec resources, and this leads to various issues.

I though I'd practice a bit and implemented a selinux_port type and a semanage provider for the same which supports managing these "natively"

The code is still a work-in-progress. There are no tests, for example, since I'm not very experienced with ruby and I find the DSLs quite confusing... I did some manual testing and it seems to work as expected.

At any rate I would like some feedback before I proceed, especially on the prefetch implementation in the provider. I can't shake the feeling that there should be a better way to do it than what I currently have, but I honestly can't tell.

Using a type should fix at least #119, #164, #93 and #38

### DIFF
--- a/lib/puppet/provider/selinux_port/semanage.rb
+++ b/lib/puppet/provider/selinux_port/semanage.rb
@@ -5,11 +5,13 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
   # SELinux must be enabled. Is there a way to get a better error message?
   confine selinux: true
 
+  # custom fact, needed for fedora 24+
+  python_command = Facter.value(:selinux_semanage_is_python3) ? 'python3' : 'python'
   # current file path is lib/puppet/provider/selinux_port/semanage.rb
   # semanage_ports.py is lib/puppet_x/voxpupuli/selinux/semanage_ports.py
   PORTS_HELPER = File.expand_path('../../../../puppet_x/voxpupuli/selinux/semanage_ports.py', __FILE__)
   commands semanage: 'semanage',
-           python: 'python'
+           python: python_command
 
   mk_resource_methods
 
--- a/lib/puppet/provider/selinux_port/semanage.rb
+++ b/lib/puppet/provider/selinux_port/semanage.rb
@@ -13,6 +13,16 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
 
   mk_resource_methods
 
+  def port_range(resource_hash)
+    low = resource_hash[:low_port]
+    high = resource_hash[:high_port]
+    if low == high
+      low.to_s
+    else
+      "#{low}-#{high}"
+    end
+  end
+
   def self.parse_helper_lines(lines)
     ret = {}
     lines.each do |line|
@@ -23,7 +33,6 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
       source, context, high, low, protocol = split
       seltype = context.split(':')[2]
       port = "#{low}-#{high}"
-      port = high if high == low
       # the ports returned by the port helper are system policy first, so the provider for any local overrides
       # will come last and "win", which is the desired behaviour
       key = "#{protocol}_#{port}"
@@ -31,7 +40,8 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
         ensure: :present,
         name: key,
         seltype: seltype,
-        ports: port,
+        low_port: low,
+        high_port: high,
         protocol: protocol.to_sym,
         source: source.to_sym
       }
@@ -53,14 +63,14 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
     instances.each do |provider|
       resource = resources[provider.name]
       if resource
-        unless resource[:ports] == provider.ports && resource[:protocol] == provider.protocol || resource.purging?
+        unless resource[:low_port].to_s == provider.low_port && resource[:high_port].to_s == provider.high_port && resource[:protocol] == provider.protocol || resource.purging?
           raise Puppet::ResourceError, "Selinux_port['#{resource[:name]}']: title does not match its port and protocol, and a conflicting resource exists"
         end
         resource.provider = provider
         resource[:ensure] = :present if provider.source == :policy
       else
         resources.values.each do |res|
-          next unless res[:ports] == provider.ports && res[:protocol] == provider.protocol
+          next unless res[:low_port] == provider.low_port && res[:high_port] == provider.high_port && res[:protocol] == provider.protocol
           warning("Selinux_port['#{resource[:name]}']: title does not match format protocol_port")
           resource.provider = provider
           resource[:ensure] = :present if provider.source == :policy
@@ -70,15 +80,15 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
   end
 
   def create
-    semanage('port', '-a', '-t', @resource[:seltype], '-p', @resource[:protocol], @resource[:ports])
+    semanage('port', '-a', '-t', @resource[:seltype], '-p', @resource[:protocol], port_range(@resource))
   end
 
   def destroy
-    semanage('port', '-d', '-p', @property_hash[:protocol], @property_hash[:ports])
+    semanage('port', '-d', '-p', @property_hash[:protocol], port_range(@property_hash))
   end
 
   def seltype=(val)
-    semanage('port', '-m', '-t', val, '-p', @property_hash[:protocol], @property_hash[:ports])
+    semanage('port', '-m', '-t', val, '-p', @property_hash[:protocol], port_range(@property_hash))
   end
 
   def exists?
--- a/lib/puppet/provider/selinux_port/semanage.rb
+++ b/lib/puppet/provider/selinux_port/semanage.rb
@@ -5,41 +5,43 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
   # SELinux must be enabled. Is there a way to get a better error message?
   confine selinux: true
 
-  commands semanage: 'semanage'
+  # current file path is lib/puppet/provider/selinux_port/semanage.rb
+  # semanage_ports.py is lib/puppet_x/voxpupuli/selinux/semanage_ports.py
+  PORTS_HELPER = File.expand_path('../../../../puppet_x/voxpupuli/selinux/semanage_ports.py', __FILE__)
+  commands semanage: 'semanage',
+           python: 'python'
 
   mk_resource_methods
 
-  def self.parse_semanage_lines(lines, source)
+  def self.parse_helper_lines(lines)
     ret = {}
     lines.each do |line|
       split = line.split(%r{\s+})
-      seltype = split.shift
-      protocol = split.shift
-      # The port-range list is comma-separated
-      ports = split.map { |range| range.delete(',') }
-      # interestingly enough, it is completely valid to define a port that overlaps with a port-range,
-      # but if an existing *exact* definition exists, that causes problems
-      ports.each do |port|
-        key = "#{protocol}_#{port}"
-        ret[key] = {
-          ensure: :present,
-          name: key,
-          seltype: seltype,
-          ports: port,
-          protocol: protocol.to_sym,
-          source: source
-        }
-      end
+      # helper format is:
+      # policy system_u:object_r:hi_reserved_port_t:s0 1023 512 tcp
+      # local  system_u:object_r:hi_reserved_port_t:s0 1023 512 udp
+      source, context, high, low, protocol = split
+      seltype = context.split(':')[2]
+      port = "#{low}-#{high}"
+      port = high if high == low
+      # the ports returned by the port helper are system policy first, so the provider for any local overrides
+      # will come last and "win", which is the desired behaviour
+      key = "#{protocol}_#{port}"
+      ret[key] = {
+        ensure: :present,
+        name: key,
+        seltype: seltype,
+        ports: port,
+        protocol: protocol.to_sym,
+        source: source.to_sym
+      }
     end
     ret
   end
 
   def self.instances
     # no way to do this with one call as far as I know
-    custom = parse_semanage_lines(semanage('port', '--list', '--locallist', '--noheading').split("\n"), :local)
-    policy = parse_semanage_lines(semanage('port', '--list', '--noheading').split("\n"), :policy)
-    # get rid of duplicates, as --list without --locallist returns local customisations as well
-    policy.merge!(custom)
+    policy = parse_helper_lines(python(PORTS_HELPER).split("\n"))
     policy.values.map do |item|
       new(item)
     end
--- a/lib/puppet/provider/selinux_port/semanage.rb
+++ b/lib/puppet/provider/selinux_port/semanage.rb
@@ -1,0 +1,85 @@
+Puppet::Type.type(:selinux_port).provide(:semanage) do
+  desc 'Support managing SELinux custom port definitions via semanage'
+
+  defaultfor kernel: 'Linux'
+  # SELinux must be enabled. Is there a way to get a better error message?
+  confine selinux: true
+
+  commands semanage: 'semanage'
+
+  mk_resource_methods
+
+  def self.parse_semanage_lines(lines, source)
+    ret = {}
+    lines.each do |line|
+      split = line.split(%r{\s+})
+      context = split.shift
+      protocol = split.shift
+      # The port-range list is comma-separated
+      ports = split.map { |range| range.delete(',') }
+      # interestingly enough, it is completely valid to define a port that overlaps with a port-range,
+      # but if an existing *exact* definition exists, that causes problems
+      ports.each do |port|
+        key = "#{protocol}_#{port}"
+        ret[key] = {
+          ensure: :present,
+          name: key,
+          context: context,
+          ports: port,
+          protocol: protocol.to_sym,
+          source: source
+        }
+      end
+    end
+    ret
+  end
+
+  def self.instances
+    # no way to do this with one call as far as I know
+    custom = parse_semanage_lines(semanage('port', '--list', '--locallist', '--noheading').split("\n"), :local)
+    policy = parse_semanage_lines(semanage('port', '--list', '--noheading').split("\n"), :policy)
+    # get rid of duplicates, as --list without --locallist returns local customisations as well
+    policy.merge!(custom)
+    policy.values.map do |item|
+      new(item)
+    end
+  end
+
+  def self.prefetch(resources)
+    # is there a better way to do this? map port/protocol pairs to the provider regardless of the title
+    # and make sure all system resources have ensure => :present so that we don't try to remove them
+    instances.each do |provider|
+      resource = resources[provider.name]
+      if resource
+        unless resource[:ports] == provider.ports && resource[:protocol] == provider.protocol || resource.purging?
+          raise Puppet::ResourceError, "Selinux_port['#{resource[:name]}']: title does not match its port and protocol, and a conflicting resource exists"
+        end
+        resource.provider = provider
+        resource[:ensure] = :present if provider.source == :policy
+      else
+        resources.values.each do |res|
+          next unless res[:ports] == provider.ports && res[:protocol] == provider.protocol
+          warning("Selinux_port['#{resource[:name]}']: title does not match format protocol_port")
+          resource.provider = provider
+          resource[:ensure] = :present if provider.source == :policy
+        end
+      end
+    end
+  end
+
+  def create
+    semanage('port', '-a', '-t', @resource[:context], '-p', @resource[:protocol], @resource[:ports])
+  end
+
+  def destroy
+    semanage('port', '-d', '-p', @property_hash[:protocol], @property_hash[:ports])
+  end
+
+  def context=(val)
+    semanage('port', '-m', '-t', val, '-p', @property_hash[:protocol], @property_hash[:ports])
+  end
+
+  def exists?
+    @property_hash[:ensure] == :present
+  end
+end
--- a/lib/puppet/provider/selinux_port/semanage.rb
+++ b/lib/puppet/provider/selinux_port/semanage.rb
@@ -13,7 +13,7 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
     ret = {}
     lines.each do |line|
       split = line.split(%r{\s+})
-      context = split.shift
+      seltype = split.shift
       protocol = split.shift
       # The port-range list is comma-separated
       ports = split.map { |range| range.delete(',') }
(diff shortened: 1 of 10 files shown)


bastelfreak:
thanks for the PR @oranenj! @petems @ffrank you are smart, do you have any suggestions for proper tests?

oranenj:
I caught one error while trying to get beaker tests to work, and apparently accidentally pushed my Gemfile hacks to get it working. hmh.

Anyway, beaker test on CentOS7 succeeded.

manifests/port.pp
```diff
@@ -12,15 +12,16 @@
...
-# @param port An network port number, like '8514'
-# @param argument An argument for semanage port. Default: "-a"
+# @param protocol Either tcp or udp.
+# @param port An network port number, like '8514', or a range like '1234-4321'
+# @param argument *deprecated*, ignored
 #
 define selinux::port (
   $context,
   $port,
+  $ensure = 'present',
```
vinzent: Are we Puppet4 only now?  If yes I would suggest starting using types now.

```
String $context,
Variant[Integer[1], Array[Integer[1],2,2]] $port,
Enum['present', 'absent'] $ensure = 'present',
Optional[Enum['tcp', 'udp', 'ipv4', 'ipv6']] $protocol = undef,
Optional[String] $argument = undef,
```

`Variant[Integer[1,65535], Array[Integer[1,65535],2,2]]` should allow to pass `port => 1` or `port => [1,34]` for a range. But I really don't know what semanage accepts as `port_range` - I could not find a specification of this.

manifests/port.pp
```diff
@@ -29,22 +30,11 @@
...
-    $port_exec_command = "add_${context}_${port}"
-  }
-
-  exec { $port_exec_command:
-    command => shellquote('semanage', 'port', $argument, '-t', $context, $protocol_switch, "${port}"), # lint:ignore:only_variable_string port can be number and we need to force it to be string for shellquote
-    # This works because there seems to be more than one space after protocol and before first port
-    unless  => sprintf('semanage port -l | grep -E %s', shellquote("^${context}  *${protocol_check}.* ${port}(\$|,)")),
-    path    => '/bin:/sbin:/usr/bin:/usr/sbin',
-    require => Class['selinux::package'],
+  validate_re($protocol, ['^tcp6?$', '^udp6?$'])
```
vinzent: is tcp6 or udp6 valid? `semanage port --help` on Fedora25 only shows udp, tcp, ipv4, ipv6.

But anyway the validate would be obsolete if using the data types.

lib/puppet/provider/selinux_port/semanage.rb
```diff
@@ -0,0 +1,77 @@
...
+          source: source
+        }
+      end
+    end
+    ret
+  end
+
+  def self.instances
+    # no way to do this with one call as far as I know
+    custom = parse_semanage_lines(semanage('port', '-lC', '-n').split("\n"), 'custom')
```
vinzent: I think the term `custom` should be called `local`. all the config-files written be semanage have the `*.local` suffix.

Caution: `semanage port` operations fail when selinux is disabled. You can add a port with `--noreload` but `--list` fails.

lib/puppet/provider/selinux_port/semanage.rb
```diff
@@ -0,0 +1,77 @@
...
+    policy = parse_semanage_lines(semanage('port', '-l', '-n').split("\n"), 'policy')
```
vinzent: I would prefer to use the options long names for a easier understanding.

manifests/port.pp
```diff
@@ -29,22 +30,11 @@
   Selinux::Port[$title] ->
```
vinzent: Ordering needs to be extended for the `absent` case

```
if $ensure == 'present' {
  Anchor['selinux::module post'] ->
  Selinux::Port[$title] ->
  Anchor['selinux::end']
} elsif $ensure == 'absent' {
  Class['selinux::config'] ->
  Selinux::Port[$title] ->
  Anchor['selinux::module pre']
} else {
  fail('Unexpected $ensure value')
}
```

bastelfreak:
@vinzent yep this is puppet4 only now

oranenj:
I fixed a few issues I found, but now I need to retire for the night...I'll address the feedback tomorrow.

manifests/port.pp
oranenj: semanage seems to accept "port" or "lower-upper" as the port definition like in the -l output. I honestly would like to keep the type as a String (or Variant[Integer, String] to reflect this as it is not immediately obvious that an array represents a range instead of individual ports, and it matches semanage output.

oranenj:
Regarding the case where SELinux is disabled... Should we fail if there are port resources, or just do nothing? 

It might be good to simply do nothing if SELinux is disabled, since that allows you to define SELinux resources without explicit checks for availability. It should work to just wrap the selinux_port resource in an if $::selinux block

lib/puppet/provider/selinux_port/semanage.rb
oranenj: Changed these to :local and symbols instead of strings

manifests/port.pp
vinzent: > I honestly would like to keep the type as a String (or Variant[Integer, String]) to reflect this 
> as it is not immediately obvious that an array represents a range instead of individual
> ports, and it matches semanage output.

Yes indeed the Array way is not so obvious as the String way, but allows to fail faster if somebody tries to pass wrong numbers.

But It can be coded  as you like. Both ways are ok for me.

vinzent:
> Regarding the case where SELinux is disabled... Should we fail if there are port resources, or just do nothing?

this is where opinions exist. :) See: https://groups.io/g/voxpupuli/topic/3891627 

In my opinion puppet-selinux should try hard to configure selinux in a way that after a reboot from disabled to enforcing all is fine without the  need to run puppet again and restart services (fine = modules loaded, booleans set, ports defined, fcontexts added, ...) and puppet should not produce puppet agent runtime errors of these resources (it may fail with an agent runtime error on setting the selinux runtime mode, but only after all other things are managed, see also #147 ).

but as semanage is IMO buggy while selinux is disabled and its therefore not so easy to do - but still doable in a hacky way.

Behaviour with current master-branch code: it just fails (ugly) in selinux disabled mode.

oranenj:
That is a very good point about not failing when SELinux goes from disabled to enabled.

I don't think we should try to hack around shortcomings in semanage. Confining the provider to systems with SELinux enabled should do just fine, and that's what I did. The error message could be better though.

vinzent:
> I don't think we should try to hack around shortcomings in semanage. Confining the provider to systems 
> with SELinux enabled should do just fine, and that's what I did. The error message could be better though.

hypothetical situation: if I have mysql on a non-standard port it will fail after the reboot. puppet needs to run , mysqld needs to be started. possibly the monitoring systems will create a bunch of incidents. different people will be involved this situation and search the problem.

`semanage port --noreload --add ...`  works just fine if selinux is disabled. the bug is that you are not able to list ports with `--list`. Interstingly it works with `--extract` (it lists all the local added/modified ports).

oranenj:
The problem is that if you want to not fail when adding a port that exists in the system policy, you need to list the system ports or you will try to add a port that exists and semanage will complain.

And I think that in such a hypothetical situation the admin should have enough foresight to boot in permissive mode. :) If you boot from disabled from enforcing, you're almost guaranteed to have all kinds of labeling issues and other things that will break the entire system anyway.

Maybe a good compromise would be to enforce permissive configuration with a warning if SELinux is disabled? In that case, the admin would see that puppet is enforcing permissive mode because SELinux is misconfigured, and upon reboot the machine will run in permissive mode until puppet executes, fixes things and sets enforcing.

vinzent:
> Maybe a good compromise would be to enforce permissive configuration with a warning if SELinux is 
> disabled? In that case, the admin would see that puppet is enforcing permissive mode because
> SELinux is misconfigured, and upon reboot the machine will run in permissive mode until puppet 
> executes, fixes things and sets enforcing.

very nice idea! downgrade enforcing to permissive while disabled and print a warning. so we don't need to care about semanage problems while in disabled mode. 👍

vinzent:
> And I think that in such a hypothetical situation the admin should have enough foresight to boot in 
> permissive mode. :) If you boot from disabled from enforcing, you're almost guaranteed to have all 
> kinds of labeling issues and other things that will break the entire system anyway.

the module creates /.autorelabel when you are in disabled mode and you wan't to go permissive/enforcing. so labels should be fine. 

but I bet  one of the 3220 downloaders of the module on forge already had this problem. :)

oranenj:
I created a pull request for that downgrade change. I'm pretty sure it should be safe.

oranenj:
I did a rebase with most of the fixes folded in to the first commit.

Save for the missing tests for the provider, I think this is starting to shape up.

lib/puppet/type/selinux_port.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  newproperty(:protocol) do
+    desc 'The protocol of the SELinux port definition'
+    newvalues(:tcp, :udp, :ipv4, :ipv6)
+    isrequired
+  end
+
+  newproperty(:context) do
```
vinzent: I think this should be renamed. context is the combination of `$selinux-user:$selinux-role:$selinux-type:$selinux-range` whereas this only is the `$selinux-type` part. 

the file resource calls the parameters `seuser`, `setype` , 'se...` maybe we should use the naming too?

vinzent: ref: https://selinuxproject.org/page/BasicConcepts

oranenj:
Partial fixes implemented. I don't have too much time today, but these are good suggestions.

oranenj:
@vinzent, do you have an opinion as to what still needs doing besides spec/acceptance tests to get this merged?

I implemented a python helper script that makes this provider *much* faster than the old semanage exec-based stuff (no-op *with purge* down from 9 seconds to 1.5 on centos 6), so I'm very interested in getting this releasable.

petems:
This is super cool! I think a few basic unit tests for the type and providers would be good, but I think we can merge this in in it's current state, with a note to get some basic unit tests added in.

Overall the code looks ok, I don't have a huge amount of knowledge of how selinux commands are run, but I think this is a good foothold to eventually move all the selinux code into providers rather than execs. It gets a +1 from me! 😄

vinzent:
@oranenj I'll have time tomorrow to run tests :-) 

Am 28. Januar 2017 15:16:53 MEZ schrieb Jarkko Oranen <notifications@github.com>:

oranenj:
I changed the port interface to support ranges explicitly, since with the new helper it's actually easy to implement. Also the port parameter no longer accepts Strings.

oranenj:
I'm aware there's quite a lot of repetition in the spec tests, but I just wanted something that works as a basic sanity check to begin with.

vinzent:
i've merged all the branches together to test it. 

for selinux::port i get: ` Error: Could not find a suitable provider for selinux_port`

```
centos-7-x64 20:35:21$ puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest.pp.Zyg9b7
  Info: Loading facts
  Warning: This method is deprecated, please use the stdlib validate_legacy function, with Stdlib::Compat::Absolute_Path. There is further documentation for validate_legacy function in the README. at ["/etc/puppetlabs/code/modules/selinux/manifests/init.pp", 61]:
     (at /etc/puppetlabs/code/environments/production/modules/stdlib/lib/puppet/functions/deprecation.rb:25:in `deprecation')
  Warning: This method is deprecated, please use the stdlib validate_legacy function, with Pattern[]. There is further documentation for validate_legacy function in the README. at ["/etc/puppetlabs/code/modules/selinux/manifests/init.pp", 62]:
     (at /etc/puppetlabs/code/environments/production/modules/stdlib/lib/puppet/functions/deprecation.rb:25:in `deprecation')
  Warning: This method is deprecated, please use the stdlib validate_legacy function, with Stdlib::Compat::String. There is further documentation for validate_legacy function in the README. at ["/etc/puppetlabs/code/modules/selinux/manifests/init.pp", 64]:
     (at /etc/puppetlabs/code/environments/production/modules/stdlib/lib/puppet/functions/deprecation.rb:25:in `deprecation')
  Warning: This method is deprecated, please use the stdlib validate_legacy function, with Stdlib::Compat::Bool. There is further documentation for validate_legacy function in the README. at ["/etc/puppetlabs/code/modules/selinux/manifests/init.pp", 65]:
     (at /etc/puppetlabs/code/environments/production/modules/stdlib/lib/puppet/functions/deprecation.rb:25:in `deprecation')
  Warning: Scope(Selinux::Port[puppet_selinux_test_policy_port_t/tcp]): [55555, 55555]
  Notice: Compiled catalog for centos-7-x64.local in environment production in 0.63 seconds
  Info: Applying configuration version '1485804923'
  Notice: /Stage[main]/Selinux::Package/Package[selinux-policy-devel]/ensure: created
  Notice: /Stage[main]/Selinux::Config/File_line[set-selinux-config-to-enforcing]/ensure: created
  Notice: /Stage[main]/Selinux::Config/Exec[change-selinux-status-to-enforcing]/returns: executed successfully
  Notice: /Stage[main]/Main/Selinux::Module[puppet_selinux_test_policy]/File[/usr/share/selinux/puppet_selinux_test_policy.te]/ensure: defined content as '{md5}900ca73aab8e8fca2f48792975935291'
  Info: /Stage[main]/Main/Selinux::Module[puppet_selinux_test_policy]/File[/usr/share/selinux/puppet_selinux_test_policy.te]: Scheduling refresh of Exec[/usr/share/selinux/puppet_selinux_test_policy.pp]
  Notice: /Stage[main]/Main/Selinux::Module[puppet_selinux_test_policy]/Exec[/usr/share/selinux/puppet_selinux_test_policy.pp]: Triggered 'refresh' from 1 events
  Notice: /Stage[main]/Main/Selinux::Module[puppet_selinux_test_policy]/Selmodule[puppet_selinux_test_policy]/ensure: created
  Notice: /Stage[main]/Main/Selinux::Boolean[puppet_selinux_test_policy_bool]/Selboolean[puppet_selinux_test_policy_bool]/value: value changed 'off' to 'on'
  Notice: /Stage[main]/Main/Selinux::Permissive[puppet_selinux_test_policy_t]/Selinux_permissive[puppet_selinux_test_policy_t]/ensure: created
  Notice: /Stage[main]/Main/Selinux::Fcontext[/tmp/fcontexts_source(/.*)?]/Selinux_fcontext[/tmp/fcontexts_source(/.*)?_a]/ensure: created
  Notice: /Stage[main]/Main/Selinux::Fcontext::Equivalence[/tmp/fcontexts_equivalent]/Selinux_fcontext_equivalence[/tmp/fcontexts_equivalent]/ensure: created
  Notice: /Stage[main]/Main/File[/tmp/fcontexts_source]/ensure: created
  Notice: /Stage[main]/Main/File[/tmp/fcontexts_equivalent]/ensure: created
  Notice: /Stage[main]/Main/File[/tmp/fcontexts_source/define_test]/ensure: created
  Info: /Stage[main]/Main/File[/tmp/fcontexts_source/define_test]: Scheduling refresh of Exec[/sbin/restorecon -FR /tmp/fcontexts_*]
  Notice: /Stage[main]/Main/File[/tmp/fcontexts_equivalent/define_test]/ensure: created
  Info: /Stage[main]/Main/File[/tmp/fcontexts_equivalent/define_test]: Scheduling refresh of Exec[/sbin/restorecon -FR /tmp/fcontexts_*]
  Notice: /Stage[main]/Main/Exec[/sbin/restorecon -FR /tmp/fcontexts_*]: Triggered 'refresh' from 2 events
  Notice: /Stage[main]/Selinux/Anchor[selinux::end]: Dependency Selinux_port[tcp_55555-55555] has failures: true
  Notice: /Stage[main]/Main/File[/tmp/test_selinux_fcontext]: Dependency Selinux_port[tcp_55555-55555] has failures: true
  Info: Class[Main]: Unscheduling all events on Class[Main]
  Info: Stage[main]: Unscheduling all events on Stage[main]
  Warning: /Stage[main]/Selinux/Anchor[selinux::end]: Skipping because of failed dependencies
  Warning: /Stage[main]/Main/File[/tmp/test_selinux_fcontext]: Skipping because of failed dependencies
  Error: Could not find a suitable provider for selinux_port
  Info: Creating state file /opt/puppetlabs/puppet/cache/state/state.yaml
  Notice: Applied catalog in 38.15 seconds
```

oranenj:
@vinzent that will happen if selinux is disabled. Is it enabled on the system you test on?

vinzent:
i've found some problem with my centos-7-x64 nodeset (leftover of some
experiment).

But nonetheless it was enabled as other tests succeeded:

Finished in 3 minutes 3.2 seconds (files took 2 minutes 26 seconds to load)
21 examples, 4 failures

Failed examples:

rspec ./spec/acceptance/class_spec.rb[1:1:1] # selinux class behaves
like a idempotent resource applies with no errors
rspec ./spec/acceptance/class_spec.rb[1:1:2] # selinux class behaves
like a idempotent resource applies a second time without changes
rspec ./spec/acceptance/class_spec.rb:91 # selinux class the test file
should have the specified file context File "/tmp/test_selinux_fcontext"
selinux_label should match /^.*:puppet_selinux_test_policy_exec_t:s0$/
rspec ./spec/acceptance/class_spec.rb:121 # selinux class port 55555
should have type puppet_selinux_test_policy_port_t Command "semanage
port -l | grep puppet_selinux_test_policy_port_t" stdout should match
/puppet_selinux_test_policy_port_t.*55555$/

vinzent:
@oranenj fixed the centos-7-x64 problem. now there is another error:

```
       	Error: /Stage[main]/Main/Resources[selinux_fcontext]: Failed to generate additional resources using 'generate': No such file or directory @ rb_sysopen - /etc/selinux/targeted/contexts/files/file_contexts.local
       	Error: Failed to apply catalog: No such file or directory @ rb_sysopen - /etc/selinux/targeted/contexts/files/file_contexts.local
```

oranenj:
I can't reproduce these errors when I run the tests separately on each branch. The beaker tests with CentOS 7.2 are all green,

vinzent:
i've pushed my test branch: https://github.com/vinzent/puppet-selinux/tree/integration_test 

- acceptan/class_spec.rb fix: https://github.com/vinzent/puppet-selinux/commit/77dc83c59f4a5559980c23e1340eaeefb8cb1832
- centos/7 beaker nodeset (centos project vagrant box instead of outdated Puppetlabs box): https://github.com/vinzent/puppet-selinux/commit/db2d9b01ab1280cc40247c93a7fe11bf5c720041

vinzent:
Am 30.01.2017 um 21:15 schrieb Jarkko Oranen:
>
> I can't reproduce these errors when I run the tests separately on each
> branch. The beaker tests with CentOS 7.2 are all green,
>

From Centos 7.2 to 7.3 selinux was replaced by a newer version. :-/
upstream box is centos 7.3. this is what i used in my second test-run:
https://github.com/vinzent/puppet-selinux/tree/integration_test)

oranenj:
I'll try with that other box. Perhaps it's wrong to assume file_contexts.local always exists.

oranenj:
At least the port provider branch seems to work okay as is.

vinzent:
the type has a problem on Fedora 25:

```
       	Info: Class[Main]: Unscheduling all events on Class[Main]
       	Info: Stage[main]: Unscheduling all events on Stage[main]
       	Warning: /Stage[main]/Selinux/Anchor[selinux::end]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Main/File[/tmp/test_selinux_fcontext]: Skipping because of failed dependencies
       	Error: Could not find a suitable provider for selinux_port
```

this failure is not present on centos-6-x64 and centos-7-64.

oranenj:
Curious. My workstation is a Fedora 25 and it worked just fine. Maybe SELinux is disabled again? Anyway, I pushed a quick fix for the first issue in the fcontext branch, though it will break spec tests.

I'll have to take a closer look later. now it's time to sleep :)

vinzent:
for Fedora 24 (`BEAKER_debug=yes BEAKER_set="fedora-24-x64" PUPPET_INSTALL_TYPE="agent" bundle exec rake beaker`) i get:

```
  Error: Failed to apply catalog: Execution of '/usr/bin/python /etc/puppetlabs/code/modules/selinux/lib/puppet_x/voxpupuli/selinux/semanage_ports.py' returned 1: Traceback (most recent call last):
    File "/etc/puppetlabs/code/modules/selinux/lib/puppet_x/voxpupuli/selinux/semanage_ports.py", line 7, in <module>
      import semanage
```

this sounds like semanage was not installed before puppet was run?

vinzent:
>  Maybe SELinux is disabled again?

right now the spec_helper_acceptance.rb takes care of setting to permissive and rebooting before starting the tests. this is because the puppetlabs boxes come with selinux=disabled . for the upstream centos/fedora boxes this would not be necessary as they come with selinux enforcing by default.

vinzent:
the python binary on Fedora 25 is python3 . A /usr/bin/python does not exist and thus the provider is not suitable because of the confine to python. :-/

if I change python to python3 in lib/puppet/provider/selinux_port/semanage.rb it works.

lib/facter/semanage_is_python3.rb
```diff
@@ -0,0 +1,6 @@
+Facter.add(:semanage_is_python3) do
```
vinzent: maybe prefix with `selinux_` to indicate that its from the selinux module?

lib/puppet/provider/selinux_port/semanage.rb
```diff
@@ -5,11 +5,13 @@
   # SELinux must be enabled. Is there a way to get a better error message?
   confine selinux: true
 
+  # custom fact, needed for fedora 24+
+  python_command = Facter.value(:semanage_is_python3) ? 'python3' : 'python'
```
vinzent: another variant woudl be

```
commands python: 'python' if Puppet::FileSystem.exist?('/usr/bin/python')
commands python: 'python3' if Puppet::FileSystem.exist?('/usr/bin/python3')
```

vinzent: forget about this. doesn't work if both pythons are installed.

oranenj: both python and python3 will be available, but only the other one might have the library needed, so I need a fact to detect this.

With this change, the acceptance tests seem to pass on everything relevant.

oranenj:
pushed a commit with a selinux_semanage_is_python3 fact

vinzent:
I think there is some debugging leftover:
```
  Warning: Scope(Selinux::Port[puppet_selinux_test_policy_port_t/tcp]): [55555, 55555]
```

vinzent:
could you do a rebase? :-)

oranenj:
All those merges were making my phone bleep non-stop :P

vinzent:
Make SELinux great again! :) thanks alot @oranenj !

PR #175, opened 2017-01-15
Downgrade enforcing to permissive configuration when SELinux is disabled

oranenj:
This avoids configurations that could break when SELinux is enabled
again. The user will be issued a warning if they are not already
enforcing permissive or disabled mode

Offshoot of pull request #174

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,11 @@ running system.
 * If SELinux is disabled and you want to switch to permissive or enforcing you
   are required to reboot the system (limitation of SELinux). The module won't
   do this for you.
-* If you add filecontexts with `semange fcontext` (what `selinux::fcontext` 
+* If SELinux is disabled and the user wants enforcing mode, the module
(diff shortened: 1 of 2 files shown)


oranenj:
Another thing this could do is ignore all the defines (which will fail when SELinux is disabled, anyway) when:
1) SELinux is disabled, but
2) The module wants to enforce either Permissive or Enforcing mode, *or*
3) The module wants to enforce disabled mode

When there are defines but SELinux mode is not explicitly configured (ie. it is undef) we can allow things to fail so that the admin has time to notice the error.

This way, you could configure SELinux resources without checking if selinux is disabled, but only if the user has explicitly set it to any of the available modes; in the case where the user has not configured SELinux and some module wants one of the resources, it will still fail with an error.

oranenj:
Though I think the above suggested change can wait until the providers are merged. It doesn't make sense to implement it in this pull request.

PR #177, opened 2017-01-15
Add selinux_fcontext and selinux_fcontext_equivalence types

oranenj:
Based on my port type, here's a pull request for fcontexts, up for review and comments.

It seemed easier to separate the equivalences from the other kind of fcontext. It might make sense to separate the defines too (ie. selinux::fcontext and selinux::fcontext::equivalence) depending on how much of the API it is okay to break.

As of the creation of this pull request, this branch breaks support for systems with old semanage versions (ie. rhel6 and older) because the accepted syntax is different. It's probably possible to subclass the provider and override the affected methods.


lib/puppet/type/selinux_fcontext.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  newproperty(:file_type) do
+    desc 'The file type to match'
+    # See semanage manual page.
+    newvalues(%r{^[abcdflps]$})
+    defaultto 'a'
+    isrequired
+  end
+
+  newproperty(:context) do
```
vinzent: same as with the selinux_port type. context is the combination of user:role:type:range https://selinuxproject.org/page/BasicConcepts 

prior art in naming the selinux parameters see the puppet file type: https://docs.puppet.com/puppet/latest/type.html#file

manifests/fcontext.pp
```diff
@@ -44,25 +44,29 @@
...
-  $restorecond_recurse = false,
+  String $pathname,
+  Enum['absent', 'present'] $ensure  = 'present',
+  Optional[String] $destination      = undef,
+  Optional[String] $context          = undef,
+  Boolean $filetype                  = false, # ignored, 
+  Optional[String] $user             = undef,
+  Optional[String] $filemode         = 'a',
+  Boolean $equals                    = false,
+  Boolean $restorecond               = true,
```
vinzent: If we are already changing things, `restorcond` with d referes to the daemon. the utility is only called `restorecon`. The parameter only runs  `restorcon` and does nothing with `restorcond`.

lib/puppet/provider/selinux_fcontext/semanage.rb
```diff
@@ -0,0 +1,87 @@
+Puppet::Type.type(:selinux_fcontext).provide(:semanage) do
+  desc 'Support managing SELinux custom fcontext definitions via semanage'
+
+  defaultfor kernel: 'Linux'
+
+  commands semanage: 'semanage'
+  confine selinux: true
```
vinzent: Leave a comment why selinux: true is used?

oranenj:
If we're really breaking APIs I might like to do away with the restorecon parameters in fcontext entirely and provide a separate helper define for restorecon execs. The API as it is is rather clunky.

oranenj:
will fix tests tomorrow

vinzent:
Am 16.01.2017 um 19:03 schrieb Jarkko Oranen:
>
> will fix tests tomorrow
>
>
no need to hurry . discussed on IRC about pushing a 1.0.0 version with
the current master and then going to 2.0.0 with all these changes. so a
user has a clear indication that that something will break instead of
pushing 0.8.0 -> 1.0.0 and everything will break.

thanks alot for your effort!

oranenj:
Latest commit attempts to restore support for old semanage. The proper way would probably be to have a custom fact query the package version, but explicit checking for RedHat <7 will probably do for most cases.

oranenj:
I don't know how to stub facter properly. I have no idea if the cause for the errors is in the spec or in the provider :/

oranenj:
I gave up on trying to get the RHEL6 specs to work, so currently the spec is actually wrong for RHEL5 and 6. It looks like the test harness just ignores facts for some reason, or maybe they are evaluated before Facter is mocked. Either way, I don't know how to fix it and my patience has been used up. 

The code itself works when I tested it, so as long as I write some acceptance tests it should be fine.

oranenj:
Currently working on extending the acceptance tests. I also simplified the fcontext and equivalence defines and removed the built-in restorecon support. I may add restorecon back at some point.

There's something weird going on with the file resource, as the files do not get created with the correct equivalences, and even though I set it up to run restorecon the contexts are still wrong until the second run...

oranenj:
Acceptance test manifest passed on CentOS6

oranenj:
CentOS7 was fine too

oranenj:
hm, now I wonder how to get the tests to pass considering the provider requires the selinux module which is not available. For the testing it is enough to mock the existence of the module, but the provider still does `require 'selinux'` which seems to not work.

vinzent:
seems like the file_context.local is not present with a default install:

```
       	Error: /Stage[main]/Main/Resources[selinux_fcontext]: Failed to generate additional resources using 'generate': No such file or directory @ rb_sysopen - /etc/selinux/targeted/contexts/files/file_contexts.local
       	Error: Failed to apply catalog: No such file or directory @ rb_sysopen - /etc/selinux/targeted/contexts/files/file_contexts.local
```

seen with centos-7-x64 (7.3), centos-6-64 (centos project upstream vagrant boxes)

vinzent:
merged your change of yesterday, centos-6 now complains about file_contexts.subs:

```
       	Error: /Stage[main]/Main/Resources[selinux_fcontext_equivalence]: Failed to generate additional resources using 'generate': No such file or directory @ rb_sysopen - /etc/selinux/targeted/contexts/files/file_contexts.subs
       	Error: Failed to apply catalog: No such file or directory @ rb_sysopen - /etc/selinux/targeted/contexts/files/file_contexts.subs
```

vinzent:
with this patch it works for me:

```
diff --git a/lib/puppet/provider/selinux_fcontext_equivalence/semanage.rb b/lib/puppet/provider/selinux_fcontext_equivalence/semanage.rb
index 9232e99..cf6613a 100644
--- a/lib/puppet/provider/selinux_fcontext_equivalence/semanage.rb
+++ b/lib/puppet/provider/selinux_fcontext_equivalence/semanage.rb
@@ -22,9 +22,10 @@ Puppet::Type.type(:selinux_fcontext_equivalence).provide(:semanage) do
   end
 
   def self.instances
-    # Allow this to fail with an exception if it does not exist
     lines = File.read(Selinux.selinux_file_context_subs_path).split("\n")
     parse_fcontext_subs_lines(lines)
+  rescue Errno::ENOENT
+    return []
   end
 
   def self.prefetch(resources)
```

oranenj:
I implemented a slightly different version of the same patch along with spec tests and rebased

PR #183, opened 2017-01-17
Selinux permissive type

### LINKED ISSUE #165: Can't specify to remove permissive domain

- Module version: 0.7.1

The defined type `selinux::permissive` does not allow to specify ensure => 'absent'.

It should remove the permissive type if possible or fail otherwise.

* Removing a module which provides a domain which is set to permissive with semanage fails.
* related to https://github.com/voxpupuli/puppet-selinux/issues/147

oranenj:
This is a rather simple provider which should fix #165. 

This provider is even backwards-compatible, I think, if adding Puppet 4 types doesn't count.

There's a small corner case documented in the README where it's non-idempotent, but I didn't think it was worth complicating the parsing logic to fix.

### DIFF
--- a/manifests/permissive.pp
+++ b/manifests/permissive.pp
@@ -1,28 +1,31 @@
 # selinux::permissive
 #
-# This method will set a context to permissive
+# This define will set an SELinux type to permissive
 #
-# @param context A particular domain-context, like "oddjob_mkhomedir_t"
+# @param seltype A particular selinux type to make permissive, like "oddjob_mkhomedir_t"
 #
 # @example Mark oddjob_mkhomedir_t permissive
-#   selinux::permissive { 'allow-oddjob_mkhomedir_t':
-#     context  => 'oddjob_mkhomedir_t',
+#   selinux::permissive { 'oddjob_mkhomedir_t':
+#     ensure => 'present'
 #   }
 #
 define selinux::permissive (
-  $context,
+  String $seltype = $title,
+  Enum['present', 'absent'] $ensure = 'present',
 ) {
 
   include ::selinux
+  if $ensure == 'present' {
+    Anchor['selinux::module post'] ->
+    Selinux::Permissive[$title] ->
+    Anchor['selinux::end']
+  } else {
+    Anchor['selinux::start'] ->
+    Selinux::Permissive[$title] ->
+    Anchor['selinux::module pre']
+  }
 
-  Anchor['selinux::module post'] ->
-  Selinux::Permissive[$title] ->
-  Anchor['selinux::end']
-
-  exec { "add_${context}":
-    command => shellquote('semanage', 'permissive', '-a', $context),
-    unless  => sprintf('semanage permissive -l | grep -Fx %s', shellquote($context)),
-    path    => '/bin:/sbin:/usr/bin:/usr/sbin',
-    require => Class['selinux::package'],
+  selinux_permissive {$seltype:
+    ensure => $ensure,
   }
 }
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ running system.
   because the used tools fail.
 * `selinux::port` has the `action` parameter which  if you specify `-d` or 
   `--delete` silently does nothing. (GH-164)
-* `selinux::permissive` allows only to set a domain to permissive but not to
-   remove it. (GH-165)
+* If you try to remove a built-in permissive type, the operation will appear to succeed
+  but will actually have no effect, making your puppet runs non-idempotent.
 
 ## Usage
 
--- a/lib/puppet/provider/selinux_permissive/semanage.rb
+++ b/lib/puppet/provider/selinux_permissive/semanage.rb
@@ -1,0 +1,63 @@
+Puppet::Type.type(:selinux_permissive).provide(:semanage) do
+  desc 'Support managing SELinux permissive types definitions via semanage'
+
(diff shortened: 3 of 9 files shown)


manifests/permissive.pp
```diff
@@ -10,7 +10,8 @@
 #   }
 #
 define selinux::permissive (
-  $context,
+  String $context,
```
vinzent: I think the param should default to `$title` if we're at it changing.

what about naming it `seltype` as in the other renewed types? I think this would be consistent.

manifests/permissive.pp
```diff
@@ -19,10 +20,7 @@
   Selinux::Permissive[$title] ->
```
vinzent: removing would need to happen before removing modules because one can not remove a module which has it set to permissive with semanage permissive.

vinzent:
added backwards-incomptabile flag because the data types are not puppet3 non-future parser compatible - and that is how most users use puppet3.

oranenj:
Something is weird with those tests I think.

I see now that it's running the context I added, but for some reason only the last bit, eg.
```
  on redhat-6-x86_64
    ensure selinux_permissive oddjob_mkhomedir_t is present
      should contain Selinux::Permissive[mycontextp] that comes before Anchor[selinux::end]
    ensure oddjob_mkhomedir_t is absent
      should contain Selinux::Permissive[mycontextp] that comes before Anchor[selinux::module pre]
```

vinzent:
> I see now that it's running the context I added, but for some reason
> only the last bit, eg.
>
> |on redhat-6-x86_64 ensure selinux_permissive oddjob_mkhomedir_t is
> present should contain Selinux::Permissive[mycontextp] that comes
> before Anchor[selinux::end] ensure oddjob_mkhomedir_t is absent should
> contain Selinux::Permissive[mycontextp] that comes before
> Anchor[selinux::module pre] |
|no actually this should print results of all tests. i get the same
result on my computer. strange.

|

vinzent:
Am 17.01.2017 um 20:52 schrieb Jarkko Oranen:
>
> Something is weird with those tests I think.
>
> I see now that it's running the context I added, but for some reason
> only the last bit, eg.
>
> |on redhat-6-x86_64 ensure selinux_permissive oddjob_mkhomedir_t is
> present should contain Selinux::Permissive[mycontextp] that comes
> before Anchor[selinux::end] ensure oddjob_mkhomedir_t is absent should
> contain Selinux::Permissive[mycontextp] that comes before
> Anchor[selinux::module pre] |
>
I suspect that selinux__permissive is now the type and the defined type
and this could confuse rspec-puppet?

oranenj:
I suspect it has something to do with all the should expressions being inside one do block.

vinzent:
> I suspect it has something to do with all the should expressions being
> inside one do block.
>
your right. doing running with dedicated `it {}` works.

oranenj:
my local rspec runs didn't give me those errors. Hm.

It's probably not kosher to reuse the let'd provider like that, but I'm not sure how to best fix it.

oranenj:
I wonder what's causing the rspec test to fail. They work on my local machine.

Maybe it's the ruby version? Hmh

oranenj:
Rebased and added some acceptance tests that pass on CentOS 6 at least. Still not sure if travis rspec will pass, though.

Warning: the acceptance tests are *super* slow because any operation with semanage seems to take ~20-30 seconds. They take about 6-7 minutes for the selinux::permissive spec

oranenj:
nil is the bane of my existence. It's not really fun to debug these test failures given that I only seem to be able to reproduce them via travis...

vinzent:
You could try to run the travis container locally: https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image

Imho travis uses ubuntu as distro.

vinzent:
tried to run with centos 5.11 :)

of course it failed. altough metadata.json states centos5 support i suspect at least selinux::module and selinux::permissive do not work at all. but this isn't related to this change.

```
  Info: /Stage[main]/Main/Selinux::Module[puppet_selinux_test_policy]/File[/usr/share/selinux/puppet_selinux_test_policy.te]: Scheduling refresh of Exec[/usr/share/selinux/puppet_selinux_test_policy.pp]
  Notice: /Stage[main]/Main/Selinux::Module[puppet_selinux_test_policy]/Exec[/usr/share/selinux/puppet_selinux_test_policy.pp]/returns: make: /usr/share/selinux/devel/Makefile: No such file or directory
```

or


```
      Failure/Error: shell('semanage permissive -a kernel_t')
      Beaker::Host::CommandFailure:
        Host 'centos-511-x64' exited with 1 running:
         semanage permissive -a kernel_t
        Last 10 lines of output were:
        		-M, --mask       Netmask
        		-P, --prefix     Prefix for home directory labeling
        		-L, --level      Default SELinux Level (MLS/MCS Systems only)
        		-R, --roles      SELinux Roles (ex: "sysadm_r staff_r")
        	
        		-s, --seuser     SELinux User Name
        		-t, --type       SELinux Type for the object
        		-r, --range      MLS/MCS Security Range (MLS/MCS Systems only)
        	
        	permissive not defined
```

vinzent:
added issue to remove centos 5 support: https://github.com/voxpupuli/puppet-selinux/issues/190

oranenj:
I fixed the beaker tests, I think.

Issue #178, opened 2017-01-17
Redesign selinux::module parameters


vinzent:
Current situation
-----------------------

* it only allows to build modules with only *.te files. no *.if and .fc extensions are possible
* it does not allow to install an already compiled module
* with some policies it's not needed to install make and co. it may install too much software not required and not wishfull on some systems.

Proposed interface
--------------------------

```
# only minimal module building (checkmodule, semodule_package)
selinux::module { 'mymodule':
  builder => 'minimal',
  source_te => 'puppet:///modules/profile/mymodule.te',
} 

# this will install additional required packages
# and build with make -f /usr/share/selinux/devel/Makefile
selinux::module { 'mymodule':
  ensure => 'present',
  builder => 'makefile', # DEFAULT 
  source_te => 'puppet:///modules/profile/mymodule.te',
  source_if => 'puppet:///modules/profile/mymodule.if',
  source_fc => 'puppet:///modules/profile/mymodule.fc',
} 

# no compiling at all, only loading an already compiled binary
selinux::module { 'mymodule':
  source_pp => 'puppet:///modules/profile/mymodule.pp',
} 

# remove a module
selinux::module { 'mymodule':
  ensure => 'absent',
} 

selinux::module { 'mymodule':
  ensure => 'disabled',
} 
```

Things to consider
-------------------------

* newer semodule utilities have the ability to specify a module priority (`-X,--priority=PRIORITY`) which isn't supported by the selmodule type. semodule defaults to priority 400 if nothing is defined (seen on EL7, Fedora 25)
* semodule provides the ability to disable and enable modules which the selmodule type does not support, valid usecase was given in #60 
* Ensure puppet agent throws errors if build or load of modules fails . It must fail every run until build and loading of the module succeeds

vinzent:
@rnelson0 and @oranenj  - your opinions?

oranenj:
Why not just a source_dir parameter? I think for the simple cases we should allow source_te perhaps as well, but having to specify everything separately is not good either, I think.

vinzent:
> Why not just a source_dir parameter? I think for the simple cases we
> should allow source_te perhaps as well, but having to specify
> everything separately is not good either, I think.
>
>
I personally have all my custom modules in 1 folder. this would sync all
my modules at once.

vinzent:
@oranenj - tried to write your proposal down as code.

```
selinux::module { 'mymodule':
  ensure => 'present',
[...]

Issue #146, opened 2016-12-23
selinux::module should not manage files in /usr


vinzent:
## Affected Puppet, Ruby, OS and module versions/distributions

- Module version:  0.5.1-rc0

## How to reproduce (e.g Puppet code you use)

add a `selinux::module` resource

## What are you seeing

the resource creates/manages a file in `/usr/share/selinux`. According FSH standard `/usr` is *Secondary hierarchy for read-only user data;*. 

In `/usr` (excluding `/usr/local`) I only expect files managed by the systems package manager.

https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard#Directory_structure

## What behaviour did you expect instead

The puppet managed file should be placed somewhere in `/var`. Maybe in some folder in puppets `$vardir` or whatever best practice is.

## references

* Depends on #178 (Redesign selinux::module parameters)

oranenj:
manifests have access to the $settings::vardir variable, which is the vardir option of the agent. For example on my desktop:
`sudo puppet apply -e 'notice($settings::vardir)'` results in `Notice: Scope(Class[main]): /var/lib/puppet` and without sudo in `Notice: Scope(Class[main]): /home/oranenj/.puppet/cache`

vinzent:
$settings::vardir is the vardir of the puppet server. Not the one of the puppet client. Unfortunatly. 

Puppet apply behaves differently if you run as root or as user.

oranenj:
Ah, yes, I misread the documentation... I wonder if there's a good way to get the configured vardir value somehow via Puppet's ruby API... `Puppet.settings.setting('vardir').value` returns nil if called straight from ruby.

oranenj:
However, it seems to work when called in a fact during a puppet run:
```
require 'puppet'
Facter.add(:puppet_agent_vardir) do
  setcode do
    Puppet.settings['vardir']
  end
end
```
```
$ FACTERLIB=. sudo -E puppet apply -e'notice($::puppet_agent_vardir)'
Notice: Scope(Class[main]): /var/lib/puppet
```

vinzent:
I think Puppet.settings outside of a puppet run needs to call some initialization method first. 

I don't think we should rely on a client provided directory. The worst case could be that somebody sets is to "/" and we're purging "/" in the end.

oranenj:
[...]

PR #189, opened 2017-01-29
Refactor selinux::module

oranenj:
I thought I'd actually implement part of what's being discussed in #178 

This doesn't yet support building the things without selinux-devel,
but it does work, and handles failures gracefully.

It doesn't look like module building needs its own provider, but behaviour on RHEL5/6 should be tested.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -11,24 +11,18 @@
 # @param type See main class
 # @param manage_package See main class
 # @param package_name See main class
-# @param sx_mod_dir See main class
+# @param module_build_root See main class
 #
 class selinux::config (
-  $mode           = $::selinux::mode,
-  $type           = $::selinux::type,
-  $sx_mod_dir     = $::selinux::sx_mod_dir,
-  $manage_package = $::selinux::manage_package,
-  $package_name   = $::selinux::package_name,
+  $mode              = $::selinux::mode,
+  $type              = $::selinux::type,
+  $manage_package    = $::selinux::manage_package,
+  $package_name      = $::selinux::package_name,
+  $module_build_root = $::selinux::module_build_root
 ) {
 
   if $caller_module_name != $module_name {
     fail("Use of private class ${name} by ${caller_module_name}")
-  }
-
-  file { $sx_mod_dir:
-    ensure => directory,
-    owner  => 'root',
-    group  => 'root',
   }
 
   if ($mode == 'enforcing' and !$::selinux) {
@@ -85,5 +79,35 @@ class selinux::config (
       line  => "SELINUXTYPE=${type}",
       match => '^SELINUXTYPE=\w+',
     }
+  }
+
+  # Module build config:
+  validate_absolute_path($module_build_root)
+
+  file {$module_build_root:
+    ensure => 'directory',
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0644',
+  }
+
+  # put helper in place:
+  file {"${module_build_root}/modules/selinux_build_module.sh":
+    ensure => 'present',
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0755',
+    source => "puppet:///modules/${module_name}/selinux_build_module.sh",
+  }
+
+  $module_build_dir = "${module_build_root}/modules"
+
+  file {$module_build_dir:
+    ensure  => 'directory',
+    owner   => 'root',
+    group   => 'root',
+    recurse => true,
+    purge   => true,
+    force   => true,
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -25,7 +25,7 @@
 # refpolicy module builder
 #   Default value: OS dependent (see params.pp)
 # @param default_builder which builder to use by default with selinux::module
-#   Default value: refpolicy
+#   Default value: simple
 # @param boolean Hash of selinux::boolean resource parameters
 # @param fcontext Hash of selinux::fcontext resource parameters
 # @param module Hash of selinux::module resource parameters
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -33,14 +33,14 @@
 # @param port Hash of selinux::port resource parameters
 #
 class selinux (
-  $mode                   = $::selinux::params::mode,
-  $type                   = $::selinux::params::type,
-  $refpolicy_makefile     = $::selinux::params::refpolicy_makefile,
-  $manage_package         = $::selinux::params::manage_package,
-  $package_name           = $::selinux::params::package_name,
-  $refpolicy_package_name = $::selinux::params::refpolicy_package_name,
-  $module_build_root      = $::selinux::params::module_build_root,
-  $default_builder        = 'simple',
+  Optional[Enum['enforcing', 'permissive', 'disabled']] $mode = $::selinux::params::mode,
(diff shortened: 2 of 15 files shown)


oranenj:
tests will be failing and there are lots of hardcoded paths for now. Still wondering what the best way to select the build dir would be. A custom fact as discussed in #146 ?

oranenj:
I implemented simple module building using just checkmodule with a helper script. Also added the custom fact. works for me, but there's still some work to do.

I noticed that the module installs the wrong packages by default. On some distros, it only installs policycoreutils, and on others it only installs selinux-policy-devel

If we really want to support modules without selinux-policy-devel etc, that needs to be refactored too.

oranenj:
I cleaned this up a bit and fixed the tests.

Acceptance goes through with CentOS6, CentOS7 and Fedora 25

oranenj:
RHEL7 default package was selinux-policy-devel too. Changed it to policycoreutils-python which is where semanage resides

oranenj:
rebased once more to clean things up. I'm pretty happy with this as it is. Any comments?

manifests/config.pp
```diff
@@ -86,4 +80,34 @@
...
+
+  file {$module_build_root:
+    ensure => 'directory',
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0644',
+  }
+
+  # put helper in place:
+  file {"${module_build_root}/modules/selinux_build_module.sh":
```
vinzent: I think this file needs `bin_t`seltype or puppet agent will maybe throw SELinux AVC's when it executes?

vinzent: maybe we should review seltypes of the directory.

manifests/init.pp
```diff
@@ -14,29 +14,33 @@
...
-# @param manage_package manage the package for selinux tools
+# @param manage_package manage the package for selinux tools and refpolicy
 #   Default value: true
 # @param package_name sets the name for the selinux tools package
 #   Default value: OS dependent (see params.pp)
+# @param refpolicy_package_name sets the name for the refpolicy development package, required for the
+# refpolicy module builder
+#   Default value: OS dependent (see params.pp)
+# @param default_builder which builder to use by default with selinux::module
+#   Default value: refpolicy
```
vinzent: default is simple isn't it?

oranenj: This is a leftover when I had refpolicy as the default earlier, but I think simple is better because it has fewer dependencies.

manifests/config.pp
oranenj: I didn't get any SELinux errors running the module, and I run my desktop in enforcing mode, so it should be fine. The acceptance tests also seemed fine.

vinzent: even if the puppet agent process runs as `puppetagent_t` type it does not produce an AVC when it executes the selinux_build_module.sh script which has type  `usr_t`. so this is fine. I just wonder why it does not.

by default the puppet agent process even does run in `unconfinged_service_t` (or `initrc_`t) because the paths in the refpolicy were not adapted to the /opt/puppetlabs change.

manifests/init.pp
```diff
@@ -14,29 +14,33 @@
...
-  $manage_package = $::selinux::params::manage_package,
-  $package_name   = $::selinux::params::package_name,
+  $mode                   = $::selinux::params::mode,
+  $type                   = $::selinux::params::type,
+  $makefile               = $::selinux::params::refpolicy_makefile,
+  $manage_package         = $::selinux::params::manage_package,
+  $package_name           = $::selinux::params::package_name,
+  $refpolicy_package_name = $::selinux::params::refpolicy_package_name,
+  $module_build_root      = $::selinux::params::module_build_root,
+  $default_builder        = 'simple',
```
vinzent: it's a good choice that the default is the one with the least required packages.

manifests/module.pp
```diff
@@ -9,75 +9,114 @@
...
+  Optional[String] $source_te = undef,
+  Optional[String] $source_fc = undef,
+  Optional[String] $source_if = undef,
+  Enum['absent', 'present'] $ensure = 'present',
+  Optional[Enum['simple', 'refpolicy']] $builder = undef,
   $syncversion  = undef,
 ) {
 
+  if $builder == 'refpolicy' {
+    require ::selinux::refpolicy_package
```
vinzent: can we maybe make this `selinux::package::policy_devel`? the term refpolicy related to  the package isn't quite right because refpolicy is what Tresys provides. The RHEL/Fedora fork has large differences to the refpolicy. 

But maybe at all not important and just a personal opinion of mine. :)

manifests/module.pp
```diff
@@ -9,75 +9,114 @@
...
-    source  => $source,
-    content => $content,
+    exec { "build-module-${title}":
+      path    => '/bin:/usr/bin',
+      cwd     => $module_dir,
+      command => "${build_command} || (rm -f ${title}.pp loaded && exit 1)",
+      creates => "${module_file}.pp",
+      notify  => Exec["install-module-${title}"],
+    }
+    # we need to install the module manually because selmodule is kind of dumb. It ends up
```
vinzent: what kind of dumb? can you elaborate more? Should we file a bug on the puppetlabs jira?

spec/acceptance/class_spec.rb
```diff
@@ -15,12 +15,44 @@ class { 'selinux': mode => 'enforcing' }
         protocol => 'tcp',
       }
 
-      # with puppet4 I would use a HERE DOC to make this pretty,
-      # but with puppet3 it's not possible.
+      # just something simple I found via Google:
+      file {'/tmp/selinux_simple_policy.te':
```
vinzent: thanks for rewriting with HERE DOC. 👍

manifests/module.pp
```diff
@@ -9,75 +9,114 @@
...
 
+  if ($builder == 'simple' and $source_if != undef) {
+    fail("The simple builder does not support the 'source_if' parameter")
+  }
+
+  # let's just make doubly sure that this is an absolute path:
+  validate_absolute_path($::selinux::config::module_build_dir)
+
+  $module_dir = "${::selinux::config::module_build_dir}/${title}"
+  $module_file = "${module_dir}/${title}"
```
vinzent: If one manages multiple refpolicy style policies, lets call them A and B,  with *.if files and A depends on an interface provided by B's *.if I don't know if they require to be in the same directory so that the makefile finds the interfaces.

this path creates a directory for every module. Is this a requirement? There should IMHO never be a same named policy source file. 

With newer (>= CentOS 7.3.) the refpolicy style will fail if filename and resulting policy name are not the same.

spec/acceptance/class_spec.rb
```diff
@@ -15,12 +15,44 @@ class { 'selinux': mode => 'enforcing' }
...
+        ensure  => 'file',
+        content => @("EOF")
```
vinzent: adding 2-space indentation to the here doc would make it more readable.

manifests/module.pp
oranenj: It's because of the syncversion bug, I think. It can only check that the module is loaded, and if you recompile a module that's loaded, it won't actually detect the change. Installing it manually is a workaround for that.

oranenj: Also, the lack of an AVC might be because the files are in the libdir which contains executables anyway (ie. facts and such). The policy seems to allow puppetagent_t to do quite a lot, so might be that it just has permission to execute whatever is in the libdir

PR #193, opened 2017-01-31
Remove Amazon Linux support

vinzent:
We don't have the ability to test Amazone Linux. Amazon Linux
is similiar to CentOS or RHEL but it's not the same.

We only list systems we test with beaker acceptance tests
as supported.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -39,17 +39,6 @@ class selinux::params {
             }
           }
         }
-        'Amazon': {
-          $sx_fs_mount = '/selinux'
-          case $os_maj_release {
-            '4': {
-              $package_name = 'policycoreutils-python'
-            }
-            default: {
-              fail("${::operatingsystem}-${::os_maj_release} is not supported")
-            }
-          }
-        }
         default: {
           case $os_maj_release {
             '7': {
--- a/metadata.json
+++ b/metadata.json
@@ -21,13 +21,6 @@
         "24",
         "25"
       ]
-    },
-    {
-      "operatingsystem": "Amazon",
-      "operatingsystemrelease": [
-        "6.0",
-        "7.0"
-      ]
     }
   ],
   "dependencies": [


alexjfisher:
There's specific support for amazon in params.pp
https://github.com/voxpupuli/puppet-selinux/blob/f3a86c66dfd192db4c191f60dc84484d07ad67bf/manifests/params.pp#L42

vinzent:
@seefood we're discussing about metadata.json and amazon linux support. 

we think for listing amazon linux we need to have acceptance tests and somebody maintaining the nodesets and/or vagrant boxes required to run tests. 

Would you be able to help us here?

vinzent:
I tried to run the tests with the Amazon Linux 2016.09.0 vagrant box (mvbcoding/awslinux). 

It wasn't able to install puppet-agent because the amazon linux yum repos were not accessible (access denied).

jyaworski:
Travis doesn't have Amazon as a bistro option?

vinzent:
beaker fails to setup the Amazon Linux host because root login is prevented with a forced command in root ssh-authorized keys. filed a ticket: https://tickets.puppetlabs.com/browse/BKR-1052

vinzent:
@jyaworski how comes travis into play when trying to run selinux acceptance tests (only possible on real systems not inside containers)?

vinzent:
got beaker working with ec2 amazon linux. params.pp fails to detect correct settings because osfamily shows RedHat not Amazon:

```
[root@ec2-35-157-122-195 ~]# facter osfamily
RedHat
[root@ec2-35-157-122-195 ~]# facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "Amazon",
  release => {
    full => "2016.09",
    major => "2016",
    minor => "09"
  },
  selinux => {
    enabled => false
  }
}
[root@ec2-35-157-122-195 ~]# facter operatingsystemmajrelease
2016
```

maybe it worked with puppet3.

vinzent:
I was wrong again. it checks for version 4 wich fails.

alexjfisher:
Facter behaves differently between the version shipped with puppet 3 and puppet 4?

https://docs.puppet.com/facter/2.4/core_facts.html#operatingsystemrelease
> On Amazon Linux, returns the lsbdistrelease value.

https://docs.puppet.com/facter/3.5/core_facts.html#operatingsystemrelease
> Linux: parse the contents of release files in /etc to retrieve the OS release.

alexjfisher:
'4' might have been the major kernel version being returned??

Regardless, the metadata.json is definitely wrong as there's no such thing as Amazon Linux 6.0 and 7.0? See https://github.com/puppetlabs/puppetlabs-ntp/blob/dccfbc8f6662b90694777c5d221f904bdb94adc0/metadata.json#L96 for some valid versions.

vinzent:
params.pp is a complete mess IMHO. :-( It selects different packages for different distro releases and I don't see an obvious reasons.

vinzent:
how should this work with amazon????

```
Feb  1 21:15:02 ip-172-31-25-242 kernel: [    0.000000] Command line: root=LABEL=/ console=tty1 console=ttyS0 selinux=0
Feb  1 21:15:02 ip-172-31-25-242 kernel: [    0.000000] Kernel command line: root=LABEL=/ console=tty1 console=ttyS0 selinux=0
```

removed selinux=0 from /boot/grub/grub.conf and /boot/grub/menu.lst and after a reboot it's still the same. VM is HVM type.

vinzent:
I give up.  If nobody steps in to provide information about howto enable selinux on amazon linux I'm for removing it. 

google reveals not much about selinux and amazon linux. I don't think there is a wide spreaded usage of selinux (if it is possible to enable it at all with a recent Amazon Linux).

jyaworski:
Pinging @rnelson0 who may have knowledge here.

alexjfisher:
@vinzent Given that the current versions of 'Amazon Linux' in metadata.json are (AFAICT) nonsense, I'm fine with removing them (this PR).  In general, working out how to do acceptance testing on Amazon Linux would be a good thing though.  Are there any other voxpupuli module's that claim Amazon Linux support?

vinzent:
> In general, working out how to do acceptance testing on Amazon
> Linux would be a good thing though.  Are there any other voxpupuli
>module's that claim Amazon Linux support?

I will provide a pr to modulesync config with ec2 beaker config. It works just not selinux particularly.

seefood:
Sorry for getting back only now.
Amazon Linux may have started as CentOS 6 -ish but it's a rolling distro (no numbered releases, just snapshots of accumulated updates). it is therefore still stuck with old package name conventions and paths, plus oddities of their own (multiple parallel versions of python, ruby, perl, etc). the kernel is 4.4 and libc is at CentOS/RHEL 7 level, but it's still not switched to systemd (and may never switch). it is therefore awkward to use third-party RPM repos (I have to use EPEL and puppetlabs for RHEL6 with it).

Also, yes: facter 2 and 3 give different return values for its name and versions. It's confusing and makes like annoying but it's not Amazon's fault.

As for whether I'm willing to maintain this or pass - Well, I have given notice at my current employer, and I'm not sure my replacement will take over this since we only have 3 AmaLinux machines left and they will be replaced within a few months to, so I don't see much chance of that. Can't help you much.

As for testing - tiny test machines can be free on Amazon (no AmaLinux installer in the wild, you either use their image on their systems or rip it manually to an offline VM). I recommended my bosses to get off that distro because of incompatibilities with puppet modules, RPM repos and VM images on other IaaS services (We also use Azure for development and test junk machines).

Bottom line, if I was the only one that needed it till today and no one will take over it, might as well drop it.

juniorsysadmin:
+1 for dropping support, I can't see the number of users using a Puppet selinux module to manage Amazon Linux instances being high

juniorsysadmin:
@alexjfisher puppet-nodejs pretends to support Amazon Linux, with one or two PRs raised indicating some level of actual usage on that platform

vinzent:
@juniorsysadmin  @alexjfisher created the PR with EC2 example beaker amazon linux nodeset: https://github.com/voxpupuli/modulesync_config/pull/314

vinzent:
@seefood 

> As for testing - tiny test machines can be free on Amazon (no AmaLinux installer in the wild, you either use their image on their systems or rip it manually to an offline VM).

I've manage to run beaker against ec2. but I was not able to enable selinux on the machine. Maybe they disabled it in the kernel.

dhollinger:
Looks good to me

PR #195, opened 2017-02-08
Redesign selinux::module defined type

### LINKED ISSUE #146: selinux::module should not manage files in /usr

### LINKED ISSUE #178: Redesign selinux::module parameters

vinzent:
Implements whats discussed in  #178. 

* Adds ability to specify a builder (simple or refpolicy). Simple does not require the make utility. 
* Stops building in /usr - uses puppet agents vardir/puppet-selinux/

Extends and closes #189 
Closes #178 
Closes #146


oranenj:
seems fine to me. Does the single-dir-building still work idempotently and correctly remove the sources when puppet stops managing a resource?

manifests/module.pp
```diff
@@ -44,11 +53,11 @@
   validate_absolute_path($::selinux::config::module_build_dir)
   validate_absolute_path($::selinux::refpolicy_makefile)
 
-  $module_dir = "${::selinux::config::module_build_dir}/${title}"
+  $module_dir = $::selinux::config::module_build_dir
```
oranenj: this and the above probably should be `$::selinux::module_build_dir` if the below is as well

manifests/module.pp
```diff
@@ -44,11 +53,11 @@
...
   $module_file = "${module_dir}/${title}"
```
oranenj: `$::selinux::config::module_build_dir` should be `$::selinux::module_build_dir`?

oranenj: There's another instance in validate_absolute_path that didn't get highlighted

vinzent: `::selinux::config::module_build_dir` != `::selinux::module_build_root` and `::selinux::module_build_dir` does not exist.

you created this var here: https://github.com/voxpupuli/puppet-selinux/pull/189/files#diff-ca85ef38c9b99f20d2d1c8d646a10e77R47 - it confused me too know. maybe we could come up with something more intuitive?  ;-)

vinzent:
> Does the single-dir-building still work idempotently and correctly remove the sources when puppet stops managing a resource?

I suspect a problem when you manage policy A with .te and .fc and then update the manifest to only manage the .te. Then the .fc file will not get removed.

manifests/module.pp
oranenj: ah, yeah, the naming is too close together... Hm. is module_source_dir any better?

oranenj:
@vinzent, do you think there's still something this needs, or could this be merged? I think the interface is fine as it is at least for my usecases, and also allows for backwards-compatible extension later on if needed.

vinzent:
> @vinzent <https://github.com/vinzent>, do you think there's still
> something this needs
I think it can be merged.

I've  squashed the commits to one for you and one of mine. so history
looks better. what do you think?

oranenj:
@vinzent: looks good to me.

vinzent:
added a small fix for the selinux_module_refpolicy acceptance test. Upcoming selinux tools 2.6 (in f26) will not allow to grant usr_t some permissions (which is a good check). 

Re-run the acceptance tests on CentOS6, CentOS7, Fedora 24 and Fedora 25. All green.

oranenj:
I added a duplicate label to get this out of the changelog in favour of the issue

PR #201, opened 2017-03-03
replace all validate functions with datatypes

bastelfreak:
bumping stdlib to 4.13.1 was also requiered, to get the absolutepath
datatype


vinzent:
@oranenj now compatability with 3.8 future parser is gone.

oranenj:
@vinzent too bad. Maybe I'll maintain a backport once 1.0.0 gets released, depending on how much work it turns out to be, but oh well...

PR #205, opened 2017-03-25
Add a convenience wrapper for restorecon execs

oranenj:
This wraps a restorecon exec so that it will execute after all other SELinux changes have been applied

Not yet very thorougly tested (or integrated with the fcontext define) but it should be useful enough standalone. Any comments?

### DIFF
--- a/manifests/exec_restorecon.pp
+++ b/manifests/exec_restorecon.pp
@@ -1,0 +1,37 @@
+# selinux::exec_restorecon
+# 
+# A convenience wrapper around a restorecon exec
+#
+# Will execute after all other SELinux changes have been applied, but before
+# Anchor['selinux::end']
+# 
+# @param path The path to run restorecon on. Defaults to resource title.
+# @param recurse Whether restorecon should recurse. Defaults to true
+# @param refreshonly see the Exec resource
+# @param unless see the Exec resource
+# @param onlyif see the Exec resource
+#
+define selinux::exec_restorecon(
+  Stdlib::Absolutepath $path        = $title,
+  Boolean              $refreshonly = true,
+  Boolean              $recurse     = true,
+  Optional[String]     $unless      = undef,
+  Optional[String]     $onlyif      = undef,
+) {
+  include ::selinux
+  $command = $recurse ? {
+    true  => 'restorecon -R',
+    false => 'restorecon',
+  }
+
+  exec {"selinux::exec_restorecon ${path}":
+    path        => '/sbin:/usr/sbin',
+    command     => sprintf('%s %s', $command, shellquote($path)),
+    refreshonly => $refreshonly,
+    unless      => $unless,
+    onlyif      => $onlyif,
+    before      => Anchor['selinux::end'],
+  }
+
+  Anchor['selinux::module post']  -> Exec["selinux::exec_restorecon ${path}"]
+}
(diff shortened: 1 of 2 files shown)


manifests/exec_restorecon.pp
```diff
@@ -0,0 +1,37 @@
...
+  Optional[String] $onlyif = undef,
+) {
+  include ::selinux
+  $command = $recurse ? {
+    true  => 'restorecon -R',
+    false => 'restorecon',
+  }
+
+  exec {"selinux::exec_restorecon ${path}":
+    path        => '/sbin:/usr/sbin',
```
vinzent: everytime i see path set on exec in forge modules i start thinking what is right: setting it here or leave it up to the user to set global defaults?

but this will work on all our supported platforms.

manifests/exec_restorecon.pp
```diff
@@ -0,0 +1,37 @@
...
+    command     => sprintf('%s %s', $command, shellquote($path)),
+    refreshonly => $refreshonly,
```
vinzent: it's refershonly ... how does it receive notifys? 

Is it tought a user defines explicit notifies?

spec/defines/selinux_exec_restorecon_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+describe 'selinux::exec_restorecon' do
```
vinzent: <3 tests. :) 👍

manifests/exec_restorecon.pp
oranenj: If you notify the define, I think all resources inside the define receive a notify too. That will work in this case.

manifests/exec_restorecon.pp
oranenj: I think this should be fine. If there's a need to fix it, that can be done without breaking things.

I like specifying paths explicitly so that the exec is not vulnerable to the user's environment being somehow compromised.

manifests/exec_restorecon.pp
oranenj: And yeah, this is not integrated with the fcontext define yet. That's trickier to do, so for now it's just a convenience wrapper for the user's own manifests.

We can later extend the fcontext define to include an optional automatic restorecon exec

manifests/exec_restorecon.pp
vinzent: yes, all resources inside receive the notify. 

just wanted to know if the explicit notify was inteded. :) this is fine with me.

manifests/exec_restorecon.pp
```diff
@@ -0,0 +1,37 @@
...
+# @param refreshonly see the Exec resource
+# @param unless see the Exec resource
+# @param onlyif see the Exec resource
+#
+define selinux::exec_restorecon(
+  Stdlib::Absolutepath $path = $title,
+  Boolean $refreshonly = true,
+  Boolean $recurse = true,
+  Optional[String] $unless = undef,
+  Optional[String] $onlyif = undef,
```
bastelfreak: can you reformat this to allign the =?

The vim plugin junegunn/vim-easy-align can easily do that for you.

oranenj:
Given that it's a pretty useful convenience tool in many cases (we use a similar custom define ourselves), I'll just deem this reviewed well enough.

PR #206, opened 2017-03-25
BREAKING: Remove restorecond management support

oranenj:
I don't think this is even used on modern systems, and the value of the
manifest is dubious. Does the default config even make sense?

For 1.0.0, I think the support should be removed and reintegrated later if it's wanted, so here's a pull request that gets rid of it.


oranenj:
Rebased to fix the documentation merge conflict

oranenj:
Rebased again after lint fixes...

PR #214, opened 2017-04-20
Allow specifying module content inline

lightoze:


manifests/module.pp
```diff
@@ -44,7 +47,7 @@
     require ::selinux::refpolicy_package
   }
 
-  if ($builder == 'simple' and $source_if != undef) {
+  if ($builder == 'simple' and ($source_if != undef or $content_if != undef)) {
```
bastelfreak: This should do the same:

```
if ($builder == 'simple' and ($source_if or $content_if)) 
```

lightoze: Not exactly (think empty string). Anyway, I just copied original code style. Do you think it must be changed?

vinzent:
can you add some spec tests?

did you run the beaker tests?

lightoze:
Beaker tests revealed that pick_default from stdlib does not work as I expected - somehow $content_if_real was not undef for `selinux_module_refpolicy_spec.rb` resource `puppet_test_a`. All fixed now.

oranenj:
blah, how does this review thing work...

manifests/module.pp
oranenj: @lightoze  I like the more compact style now that there is more than one parameter, but it doesn't matter that much. Fix it if you like. :)

lightoze:
Added more documentation, also made a fix in existing example (wrong 'builder' attribute).

oranenj:
Great. Thanks!

I could use this feature myself, so I'll merge this as is and do some testing later.

PR #218, opened 2017-05-02
Fedora 26 uses the same package_name as Fedora 25

logic:
Fedora 26 is almost upon us (it's available today as an alpha release), and it uses the same package name as Fedora 25.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -34,7 +34,7 @@ class selinux::params {
             '21','22','23' : {
               $package_name = 'policycoreutils-devel'
             }
-            '24', '25' : {
+            '24','25','26' : {
               $package_name = 'policycoreutils-python-utils'
             }
             default: {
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
Would it be worth changing the match to be any Fedora version about a certain value, or is the history a good indicator that the value for 27 will change yet again?

oranenj:
I agree with changing the default. That seems sensible.

It will fail anyway, if the package name changes.

logic:
It's technically legal to install a package named `/usr/sbin/semange`, which `dnf` will resolve to whatever package provides that, and I *think* puppet's package provider will do the right thing with it, but that's not exactly a common practice (and likely to surprise users of the module). ;)

I've updated the diff to make 24+ the new default, and removed the explicit failure.

oranenj:
Thanks!

PR #224, opened 2017-06-22
puppet3 module fixes

kreon:
Hi

A few fixes for selinux module:
1. Added strict mode for centos 5/6 policy
2. Fixed the case when selinux was in permissive but is set to disabled, that causes puppet try to do setenforce 0 again and again until reboot.


oranenj:
Seems fine to me, though I'm not very much interested in maintaining the puppet3 branch, I don't see a reason not to merge fixes like these.

@vinzent: any opinions?

Issue #230, opened 2017-07-27
Amazon Linux support


clinty:
Please add support for Amazon Linux.  It should be similar to RHEL5, with
`$sx_fs_mount = '/selinux'` and `$package_name = 'policycoreutils'`

Output of `facter os`:

```
{
  architecture => "x86_64",
  distro => {
    codename => "n/a",
    description => "Amazon Linux AMI release 2017.03",
    id => "AmazonAMI",
    release => {
      full => "2017.03",
      major => "2017",
      minor => "03"
    },
    specification => ":base-4.0-amd64:base-4.0-noarch:core-4.0-amd64:core-4.0-noarch"
  },
  family => "RedHat",
  hardware => "x86_64",
  name => "Amazon",
  release => {
    full => "2017.03",
    major => "2017",
    minor => "03"
  },
  selinux => {
    config_mode => "enforcing",
    current_mode => "enforcing",
    enabled => true,
    enforced => true,
    policy_version => "30"
  }
}
```

oranenj:
Is there a way to test on Amazon Linux? IIRC the lack of good means to run tests was the main reason we dropped support during the 1.0.0 refactoring.

clinty:
I don't think I understand the question.  Are you asking for a way that doesn't require running an Amazon Linux instance on AWS?

oranenj:
@clinty: yes. Beaker tests for other operating systems can be run locally using Virtualbox.

clinty:
If someone made an unofficial virtualbox image, would that suffice?

oranenj:
A vagrant box that beaker can use would be fine, I suppose. I would accept patches for Amazon Linux even if it weren't tested, but without a way to run acceptance tests, there's no way to guarantee things will not break between releases.

clinty:
Pull request sent.  I'll pursue the testing angle.

clinty:
Does https://hub.docker.com/_/amazonlinux/ help at all?

oranenj:
@clinty: This issue was closed by the pull req merge. You should probably create another issue for the testing stuff.

Unfortunately, SELinux can't really be tested inside a container, so it'll have to be a virtual machine.

It does look like there's an amazon linux vagrant box: https://app.vagrantup.com/mvbcoding/boxes/awslinux/

That is something which might be useful.

PR #231, opened 2017-07-29
Add Amazon Linux support

clinty:
In the case of SELinux, Amazon Linux appears to resemble RHEL 5.
This allows the module to compile, rather than choke, on Amazon
Linux 2017.03.

Closes: #230

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -39,6 +39,10 @@ class selinux::params {
             }
           }
         }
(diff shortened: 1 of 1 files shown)


oranenj:
I'll merge this as is. If you can figure out a way to do testing without actually paying Amazon, we can even make it "official" in the metadata. :)

Thanks.

Issue #212, opened 2017-04-10
Expected parameter 'refpolicy_makefile' of 'Class[Selinux]' to have type Stdlib::Absolutepath


nmaludy:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 
```
$ puppet --verison
3.8.6
```

- Ruby:
```
$ ruby --version
ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
```

- Distribution:
```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.3 (Maipo)
```

- Module version:
```
$ cat /etc/puppet/environments/production/modules/selinux/metadata.json | head -n 3
{
  "name": "puppet-selinux",
  "version": "1.0.0",
```

## How to reproduce (e.g Puppet code you use)
Update to puppet-selinux v1.0.0 and run the simple manifest:
```
#
class encore_rp::profile::selinux {
  class { '::selinux':
    mode => 'permissive',
  }
}
```

## What are you seeing
The following error occurs when running `puppet agent -t `:

```
[root@nickpuppet6 ~]# puppet agent -t
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Expected parameter 'refpolicy_makefile' of 'Class[Selinux]' to have type Stdlib::Absolutepath, got String at /etc/puppet/environments/production/modules/encore_rp/manifests/profile/selinux.pp:3 on node xxx
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

## What behaviour did you expect instead
The documented "Basic usage" in the readme to work without error.

oranenj:
The selinux module now uses puppet types for validation, which requires Puppet 4.

I will close this issue, as Puppet 3.8.6 is not supported by 1.0.0 anymore. Puppet 3 is EOL, and I recommend you upgrade to Puppet 4 if at all possible. It's much better in addition to being actually supported.

If you can use the future parser, you can make the module 3.8.7 future parser -compatible by replacing all mentions of  `Stdlib::Absolutepath` with `String`. This makes the module a bit more brittle since the parameters won't be validated as paths, but it'll work.

oranenj:
[...]

PR #236, opened 2017-09-17
release 1.3.0

bastelfreak:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,27 @@
-# Change log
+# Changelog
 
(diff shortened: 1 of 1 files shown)


ekohl:
Why are those issues moved around?

bastelfreak:
mhm good question. I regenerated it again and they are still missing.

wyardley:
> mhm good question. I regenerated it again and they are still missing.
https://github.com/voxpupuli/puppet-selinux/issues/212
Guessing because of labels? wontfix is now in the list of `config.exclude_labels` in Rakefile

PR #242, opened 2017-11-17
release 1.4.0

bastelfreak:


bastelfreak:
@vinzent I think we should release this as a major one, since we dropped support for one Fedora release? If I get semver right, this is worth a major one.

bastelfreak:
Oh I thought I did that, one sec

wyardley:
I guess it depends if we're considering dropping support for EOL OSes breaking... in theory, the module would likely still work, and could be used, though at some point, it could of course be broken since we're not supporting it / testing it. My suggestion is maybe not flag it as breaking?

PR #246, opened 2017-12-22
Fixing change-selinux-status-to-disabled exec

bjvrielink:
setenforce only accepts enforcing or permissive (or 0 and 1) as possible modes.
When calling the selinux module with the current mode as either enforcing or permissive and $mode => disabled, the exec change-selinux-status-to-disabled will give exit code 1. A reboot does solve this, but until that happens. each puppet run will result in a report with a corrective change.

PR #247, opened 2018-01-04
modulesync 1.6.0

bastelfreak:


bastelfreak:
Fixes https://github.com/voxpupuli/puppet-selinux/pull/246

Issue #245, opened 2017-12-22
Skip exec "change-selinux-status-to-disabled" when current mode is enforcing or permissive


bjvrielink:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.9.4
- Ruby: 2.1.9p490
- Distribution: RHEL 7
- Module version: 1.5.0

```
## How to reproduce (e.g Puppet code you use)
class {'::selinux':
  mode => 'disabled',
}
```
```
$ getenforce 
Enforcing
```

## What are you seeing
Each run, the exec 'change-selinux-status-to-disabled' runs.

## What behaviour did you expect instead
The exec 'change-selinux-status-to-disabled' should not run. This is because 'setenforce disabled' is not a correct usage of the setenforce command.
```
# setenforce disabled
usage:  setenforce [ Enforcing | Permissive | 1 | 0 ]
```

bjvrielink:
See PR #246

ekohl:
Fixed in https://github.com/voxpupuli/puppet-selinux/pull/246. Thanks!

PR #249, opened 2018-01-11
Update and check runtime SELinux status correcty

weaselshit:
The #245 issue was addressed by #246, but that introduced a new problem:
when mode param is set to 'disabled', runtime mode never gets changed to
'permissive' anymore.

The original problem was different: it was not wrong parameter to
setenforce, instead, the unless command was wrong, always grepping for
'disabled|disabled'.

This change uses literal modes instead of numeric for both exec *and*
unless command and removes conditinal statement added in #246 to restore
original behaviour. It also adds tests for unless param of the exec in
question.

@bjvrielink, could you please test if this works correctly in your environment?

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -55,13 +55,13 @@ class selinux::config (
 
     case $_real_mode {
       'permissive', 'disabled': {
-        $sestatus = '0'
+        $sestatus = 'permissive'
         if $_real_mode == 'disabled' and defined('$::selinux_current_mode') and $::selinux_current_mode == 'permissive' {
           notice('A reboot is required to fully disable SELinux. SELinux will operate in Permissive mode until a reboot')
         }
       }
       'enforcing': {
-        $sestatus = '1'
+        $sestatus = 'enforcing'
       }
       default : {
         fail('You must specify a mode (enforced, permissive, or disabled) for selinux operation')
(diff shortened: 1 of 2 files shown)


bjvrielink:
When SELinux is disabled, it is not possible to use the setenforce command (it will always print SELinux is disabled and return 1).
I don't exactly understand what issue you see, but I think you are referring to the case when the current mode is enforcing and you want to change to disabled. In this case it indeed may make sense to set the runtime mode to permissive, until a reboot happens.
Neither the old version, nor the #246 version did do this. Your version does cover this case.

I'm unable to test this code at the moment (too much distance between the Internet and my environment), but the logic behind the code looks almost OK.
The only thing I 'notice' is that when $::selinux_current_mode is 'enforcing', the notice on line 60 does not happen on the first Puppet run (as current mode then is enforcing) but does on subsequent runs (as the current mode has become 'permissive'.

vinzent:
@weaselshit do I understand correctly your  situation:

* runtime-mode: enforcing
* desired-mode: disabled
* puppet-selinux does not switch runtime-mode to permissive

correct?

weaselshit:
Hi, @vinzent, sorry for late reply. You are absolutely correct.

@bjvrielink, I think you may be missing something (or is it me?)
> When SELinux is disabled, it is not possible to use the setenforce command (it will always print SELinux is disabled and return 1).

As far as I see, we should never get to running setenforce, if it's already disabled of in a desired state. https://github.com/voxpupuli/puppet-selinux/pull/249/files#diff-0ac866172efd29d0ddc2aa75ae6f182dR85 should take care of it. I may be wrong, but I think that your original problem was caused by wrong unless clause, which I try to address with this PR.

vinzent:
@weaselshit I've tracked down why the class_disabled_spec.rb acceptance test didn't discover this. 

the testing VM was in Permissive state then the `class { selinux: ensure => 'disabled',}` resource was applied. 

i'm updating the tests.

vinzent:
ran updated acceptance of #250 tests on centos6+7, fedora 26+27. all green.

PR #271, opened 2018-09-21
modulesync 2.1.0 & add puppet 6 support

bastelfreak:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'puppetlabs_spec_helper', '~> 2.10.0',                        :require => false
+  gem 'puppetlabs_spec_helper', '>= 2.10.0',                        :require => false
   gem 'rspec-puppet-facts',                                         :require => false
   gem 'rspec-puppet-utils',                                         :require => false
   gem 'puppet-lint-leading_zero-check',                             :require => false
@@ -76,7 +76,7 @@ else
   gem 'facter', :require => false, :groups => [:test]
 end
 
-ENV['PUPPET_VERSION'].nil? ? puppetversion = '~> 6.0' : puppetversion = ENV['PUPPET_VERSION'].to_s
+ENV['PUPPET_VERSION'].nil? ? puppetversion = '~> 5.0' : puppetversion = ENV['PUPPET_VERSION'].to_s
 gem 'puppet', puppetversion, :require => false, :groups => [:test]
 
 # vim: syntax=ruby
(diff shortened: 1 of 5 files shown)


metadata.json
```diff
@@ -27,12 +27,16 @@
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">= 4.13.1 < 6.0.0"
+    },
+    {
+      "name": "puppetlabs/selinux_core",
```
alexjfisher: I think this is wrong. The module is only marked as puppet >= 6 compatible.

oranenj: hm. It is a dependency though :/ Puppet's metadata format is not expressive enough in this case.

I think in order to maintain backwards compatibility, we'll just need to note in the readme that selinux_core is needed for Puppet >= 6.0.0

oranenj:
I did a bit of testing using my test puppet master.

I upgraded it to Puppet 6 and then added selinux_core to an environment containing puppet 5 agents. It seems the puppet 5 agents either ignore or do not mind that they get synced the extra providers; selmodule seems to continue working as previously.

Therefore it seems safe to include the dependency on selinux_core in the metadata file, though I'm still undecided as to whether we should.

metadata.json
alexjfisher: That *shouldn't* be necessary.  If your puppet-agent doesn't automatically come with it, something has gone wrong.

oranenj:
https://puppet.com/docs/puppet/6.0/type.html#supported-types-in-modules states that selmodule is included in the puppet 6 distribution. Therefore having it as a dependency is redundant, since it is already implied by having Puppet installed.

alexjfisher:
@oranenj What we're not 100% sure on is the implications of putting it in `.fixtures.yml`.  It appears to be required for puppet 6, and currently not break puppet 5, but it still feels wrong/dangerous loading code that isn't supposed to be there into a puppet 4/5 master's environment.  If it works today, will it work tomorrow?

I'd still like to wait a bit longer for some guidance to come back from puppet. ping @binford2k ;)

Maybe puppet will ship the core modules as gems??  Maybe they'll update puppetlabs_spec_helper??

Gemfile
```diff
@@ -11,8 +11,7 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'puppetlabs_spec_helper', '~> 2.6',                           :require => false
-  gem 'rspec-puppet', '~> 2.5',                                     :require => false
+  gem 'puppetlabs_spec_helper', '>= 2.10.0',                        :require => false
```
alexjfisher: 2.11.0 has the change needed for .fixtures

PR #280, opened 2019-02-28
Simplify parameter handling

ekohl:
Note this silently adds RHEL 8 support. Because it's not tested I haven't added it to metadata.json, but the package name matches Fedora.


oranenj:
Seems good to me. Is the selinux mount path no longer used anymore?

ekohl:
Not for a long time. 4622a188393a256111c8f7c8d4e51471f967b967 switched to using get/setenforce.

PR #281, opened 2019-02-28
Builder improvements and acceptance tests

ekohl:
From the individual commits:

* By only including the build directories when building modules, we maintain a cleaner system.
* When the builder is loaded from the default builder, we need to make sure we include the right packages.
* This module doesn't work on anything but Red Hat based systems so we can drop the Arch and EC 2 nodesets. The tests also depend on the order so this sets the state prior to running a class. Docker doesn't support selinux so the beaker-vagrant gem is added.


Gemfile
```diff
@@ -54,12 +54,14 @@ group :system_tests do
...
   gem 'beaker-hostgenerator', '>= 1.1.22',  :require => false
   gem 'beaker-docker',                      :require => false
+  gem 'beaker-vagrant',                     :require => false
   gem 'beaker-puppet',                      :require => false
   gem 'beaker-puppet_install_helper',       :require => false
   gem 'beaker-module_install_helper',       :require => false
   gem 'rbnacl', '>= 4',                     :require => false if RUBY_VERSION >= '2.2.6'
   gem 'rbnacl-libsodium',                   :require => false if RUBY_VERSION >= '2.2.6'
   gem 'bcrypt_pbkdf',                       :require => false
+  gem 'beaker-vagrant',                     :require => false
```
bastelfreak: can you remove one `beaker-vagrant` entry?

manifests/build.pp
```diff
@@ -0,0 +1,63 @@
...
+#   modules/ # module source files and compiled policies
+#   modules/tmp # repolicy tempfiles (created by scripts)
+# ```
+#
+# @param module_build_root See main class
+#
+# @api private
+#
+class selinux::build(
+  $module_build_root = $::selinux::module_build_root,
```
bastelfreak: please don't add `::` after the `$` and add a datatype

bastelfreak: crap, `::` is used all over the place. should we keep it here as well? Up to you.

ekohl: I considered and I preferred to be consistent so kept it.

Gemfile
ekohl: Rebase gone wrong. Initially I did manually and then the proper way so modulesync wouldn't be unhappy with me.

PR #282, opened 2019-02-28
modulesync 2.5.1 and drop Puppet 4

bastelfreak:
modulesync 2.5.1


ekohl:
Now has a merge conflict

spec/defines/selinux_exec_restorecon_spec.rb
```diff
@@ -2,10 +2,14 @@
 
 describe 'selinux::exec_restorecon' do
   on_supported_os.each do |os, _facts|
+
+
     context "on #{os}" do
       context 'with resource titled /opt/mycustompath' do
         let(:title) { '/opt/mycustompath' }
-
+        let :facts do
```
ekohl: You can do this at a higher level (just below `context "on #{os}" do`) and just do it once

spec/default_module_facts.yml
```diff
@@ -1,7 +1,4 @@
 ---
-osfamily: RedHat
-operatingsystem: RedHat
-operatingsystemmajrelease: '7'
 # concat facts
 id: 0
```
ekohl: This and path can also be dropped, as well as the comment

manifests/params.pp
```diff
@@ -5,7 +5,7 @@
 class selinux::params {
   $module_build_root = "${facts['puppet_vardir']}/puppet-selinux"
 
-  case $facts['osfamily'] {
+  case $facts['os']['family'] {
     'RedHat': {
       if $facts['operatingsystem'] == 'Amazon' {
```
ekohl: If you want to use the modern facts, then this should probably also be replaced (as well as the one below).

spec/classes/selinux_package_spec.rb
```diff
@@ -9,7 +9,8 @@
             osfamily: 'RedHat',
```
ekohl: You can drop the old facts here if you want.

manifests/params.pp
```diff
@@ -5,12 +5,12 @@
...
-  case $facts['osfamily'] {
+  case $facts['os']['family'] {
     'RedHat': {
-      if $facts['operatingsystem'] == 'Amazon' {
+      if $facts['os']['name'] == 'Amazon' {
         $package_name = 'policycoreutils'
       } else {
-        $package_name = $facts['operatingsystemmajrelease'] ? {
+        $package_name = $facts['os']['release']['major'] ? {
           '5'     => 'policycoreutils',
```
ekohl: Do you need to convert this to integers or are they still strings? I think the type should match in the specs.

bastelfreak: that isn't required. Also facter displays it as string, not integer.

bastelfreak:
@ekohl IMO merging this and doing a major release is fine, that's wha we agreed on in https://github.com/voxpupuli/voxpupuli.github.io/pull/157

ekohl:
@bastelfreak I'd prefer to do a release, bump major and then wait for other changes to come in

dhollinger:
Based on our conversation in https://github.com/voxpupuli/voxpupuli.github.io/pull/157, we wanted to not ever do a release just for removing puppet 4. As such, I'm ok with this PR, @ekohl. Doing a minor/patch release to prepare for a major release change that removes Puppet 4 is exactly what the community agreed to avoid doing.

PR #283, opened 2019-04-07
modulesync 2.7.0

bastelfreak:
modulesync 2.7.0


ekohl:
Updated so the tests should pass. Also did some cleaning up of the fixtures while I was at it :)

Issue #288, opened 2019-05-30
Fcontext fails on re-run on newer platforms


ehelms:
An example error from the console of what the module runs:

```
# /sbin/semanage fcontext -a -t squid_cache_t -f a '/var/spool/squid(/.*)?'
ValueError: File context for /var/spool/squid(/.*)? already defined
```

Package version:

```
policycoreutils-python-utils.noarch                                           2.8-16.1.el8                                        @rhel-8-for-x86_64-baseos-rpms                                               
```

Seems that newer semanage will error if a `-a` operating is ran for an existing entity which breaks re-runs.

ehelms:
@ekohl any design thoughts on how to handle this within the provider code?

PR #289, opened 2019-05-30
Check if contexts exist and add or modify fcontext

ehelms:
This is an attempted solution at solving the issue described in -- https://github.com/voxpupuli/puppet-selinux/issues/288


ehelms:
Yes, RHEL 8 has it in the base policy not the local rules. Seems like two paths:

 a) change the prefetch to check all rules
 b) update puppet-squid (where I saw this issue) to only do the selinux part on non-EL8 or newer Fedora

ekohl:
From a technical perspective I'd say the first option is probably better for squid because if you change the cache directory, it'll still need to apply custom contexts to that directory. Knowing whether it should or should not is part of the selinux module. It is however more work and might break other use cases in puppet-selinux.

Anyone has opinions on this? @alexjfisher perhaps?

ehelms:
I changed the approach to have the prefetch logic check local and non-local for the source in question to ensure it does not attempt to create it if it already exists by default.

lib/puppet/provider/selinux_fcontext/semanage.rb
```diff
@@ -71,11 +71,17 @@ def self.instances
     # should never conflict with system policy
     # Old semanage fails with --locallist, use -C
     local_fcs = Selinux.selinux_file_context_local_path
+    fcs = Selinux.selinux_file_context_path
+
+    fcontext_lines = []
+
+    fcontext_lines = parse_fcontext_lines(File.readlines(fcs)) if File.exist? fcs
```
ekohl: I'm not a fan of this guard clause here. It's really hard to read for me. That it was already defined 2 lines above made me think that was redundant. Personally I'd prefer a classic if/else or a ternary.

Given the duplication, how about a helper:

```ruby
def parse_fcontext_file(path)
    return [] unless File.exist?(path)
    parse_fcontext_lines(File.readlines(path))
end

def self.instances
    parse_fcontext_file(Selinux.selinux_file_context_path) + parse_fcontext_file(Selinux.selinux_file_context_local_path)
end
```

ehelms:
Updated based on a few of the suggestions. THoughts?

ehelms:
@ekohl Any thoughts here?

ekohl:
> The purging is not an issue (as I thought in [#289 (review)](https://github.com/voxpupuli/puppet-selinux/pull/289#pullrequestreview-245038369)) because that's about unmanaged resources but this just loads more managed resources.

Ok, this was wrong and it actually is an issue. It will indeed try to purge all built in types where previously it would only attempt to purge the local overrides. Checking if we can avoid this somehow.

ekohl:
Closing in favor of #290 which was merged.

PR #290, opened 2019-06-14
Load system policy contexts

### LINKED ISSUE #288: Fcontext fails on re-run on newer platforms

ekohl:
This change prefetches the system policy contexts. This means that if the system policy already exists for `/var/spool/squid(/.*)?` then it won't attempt to override it when the specification matches. It does still allow modifications if for some reason the base policy is incorrect. Purging of all local contexts is also still possible.

Inspired by #289

Fixes #288


lib/puppet/provider/selinux_fcontext/semanage.rb
```diff
@@ -66,22 +66,29 @@ def self.parse_fcontext_lines(lines)
...
-    end
+    self.parse_fcontext_file(Selinux.selinux_file_context_local_path)
+  end
+
+  def self.system_policy_instances
+    self.parse_fcontext_file(Selinux.selinux_file_context_path)
   end
 
   def self.prefetch(resources)
-    # is there a better way to do this? map port/protocol pairs to the provider regardless of the title
```
ekohl: Not sure if this comment still makes sense.

PR #292, opened 2019-06-26
Change - Update module parameters to support older Fedora releases

blackknight36:
Recent changes to the manifest/params.pp file have caused breakage
on servers running older Fedora release.  Update manifest to avoid
this in environments that still need support for legacy hosts.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,21 +5,31 @@
 class selinux::params {
   $module_build_root = "${facts['puppet_vardir']}/puppet-selinux"
 
-  case $facts['os']['family'] {
-    'RedHat': {
-      if $facts['os']['name'] == 'Amazon' {
-        $package_name = 'policycoreutils'
-      } else {
-        $package_name = $facts['os']['release']['major'] ? {
-          '5'     => 'policycoreutils',
-          '6'     => 'policycoreutils-python',
-          '7'     => 'policycoreutils-python',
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -5,21 +5,31 @@
...
+      $package_name = $facts['os']['release']['major'] ? {
+        '5'     => 'policycoreutils',
+        '6'     => 'policycoreutils-python',
+        '7'     => 'policycoreutils-python',
+        default => 'policycoreutils-python-utils',
+      }
+    }
+
+    'Fedora': {
+      $sx_fs_mount = '/sys/fs/selinux'
```
ekohl: This variable is unused

manifests/params.pp
```diff
@@ -5,21 +5,31 @@
...
-          '6'     => 'policycoreutils-python',
-          '7'     => 'policycoreutils-python',
-          default => 'policycoreutils-python-utils',
-        }
+  case $facts['os']['name'] {
+    'Amazon': {
+      $package_name = 'policycoreutils'
+    }
+
+    'CentOS', 'RedHat': {
```
ekohl: You are breaking other RHEL derivatives here.

blackknight36:
@ekohl It looks like using hiera to define the package name will work for my needs.  I'm going to close this request.

PR #294, opened 2019-07-17
Update supported OS list in module metadata

blackknight36:
Added RedHat 8 and Fedora 30 to the supported OS list.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -12,21 +12,24 @@
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
         "6",
-        "7"
+        "7",
+        "8"
       ]
     },
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
         "6",
-        "7"
+        "7",
+        "8"
       ]
     },
     {
       "operatingsystem": "Fedora",
       "operatingsystemrelease": [
         "28",
-        "29"
+        "29",
+        "30"
       ]
     }
   ],


metadata.json
```diff
@@ -12,21 +12,16 @@
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
         "6",
-        "7"
-      ]
-    },
-    {
-      "operatingsystem": "CentOS",
```
ekohl: We still support CentOS. I don't see a reason to drop this.

blackknight36: Whoops, that wasn't intentional.  Should CentOS 8 be listed in the metadata?  I know it's not out yet but adding it to the OS list now would avoid having to update metadata in the future.

blackknight36: Also, Fedora 28 is now EOL.  Should that be removed?

ekohl: Yes, I think we should list CentOS 8. If there's an issue, it'd be a bug we need to fix. We can also reasonably assume that when it works on RHEL 8, it'll work on CentOS 8.

I'd leave F28 in for now because it can be seen as a breaking change and I've done enough of those for now ;) There's no additional code needed so there's little benefit.

blackknight36: Okay, I've updated this branch to include CentOS 8 along with RHEL 8 and Fedora 30.

ekohl:
Thanks!

PR #297, opened 2019-08-02
Use legacy facts in the params class

blackknight36:
Recent changes to the params class cause puppet runs to fail on
older clients that do not have an "os" fact defined.  Unit tests
in Travis-CI are also failing when other modules attempt to use
this module as a dependecy.  For example, see the following URL:

https://travis-ci.org/puppetlabs/puppetlabs-postgresql/jobs/567084563

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,12 +5,12 @@
 class selinux::params {
   $module_build_root = "${facts['puppet_vardir']}/puppet-selinux"
 
-  case $facts['os']['family'] {
+  case $facts['osfamily'] {
     'RedHat': {
-      if $facts['os']['name'] == 'Amazon' {
+      if $facts['operatingsystem'] == 'Amazon' {
         $package_name = 'policycoreutils'
       } else {
-        $package_name = $facts['os']['release']['major'] ? {
+        $package_name = $facts['operatingsystemmajrelease'] ? {
           '5'     => 'policycoreutils',
           '6'     => 'policycoreutils-python',
           '7'     => 'policycoreutils-python',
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
#282

blackknight36:
IMO I don't see any harm in continuing to support Puppet 4.  Heck, we still have hundreds of devices in production running Puppet 3.8 and I'm sure we're not the only organization out there with legacy devices.

bastelfreak:
Hi @blackknight36. Puppet 3 is end of life since a long time, Puppet 4 since January. We do not officially support operatingsystems nor Puppet versions that are end of life. This is documented in a [blog post](https://voxpupuli.org/blog/2019/01/03/dropping-puppet4/). We had a pretty long discussion about this but voted for dropping it. The structured facts should be available in Puppet 4 as well and this module might work there. But we don't support it. Also the old topscope facts will disappear in a future release of Puppet, so we won't migrate back to them.

---

The puppetlabs-postgresql tests are failing because the mocking isn't optimal. The module mocks all needed facts in a manual way:

https://github.com/puppetlabs/puppetlabs-postgresql/blob/0b04bc5dfda24f992bbe2f6a48ce7aa0142a50a4/spec/unit/classes/server/config_spec.rb#L9-L20

This often causes issues with dependent modules, as in this case.The modern way would be to use [rpsec-puppet-facts](https://github.com/mcanevet/rspec-puppet-facts#rspec-puppet-facts).

blackknight36:
@bastelfreak Thanks.  I will see what I can do with the postgresql module.

PR #298, opened 2019-09-25
Fix selinux_port on RHEL 8/CentOS 8

oranenj:
Hi,

this PR makes the selinux_port helper use the platform Python interpreter if it exists. This makes it work on on RHEL 8 / CentOS 8.

The platform python interpreter is guaranteed to exist, as several RHEL components depend on it.

This has seen very light testing, but at least it didn't break my test system.

### DIFF
--- a/lib/facter/selinux_python_command.rb
+++ b/lib/facter/selinux_python_command.rb
@@ -1,0 +1,15 @@
+Facter.add(:selinux_python_command) do
+  confine osfamily: 'RedHat'
+  setcode do
+    if File.exist? '/usr/libexec/platform-python'
+      # RHEL 8 / CentOS 8
+      '/usr/libexec/platform-python'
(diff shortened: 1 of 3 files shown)


oranenj:
@ekohl possibly, though honestly since the best Python is platform-specific it would probably make most sense as a hardcoded parameter.

I originally wrote the helper script because semanage was *way* too slow and Puppet's built-in ruby didn't have the required APIs for SELinux to do the inspection "natively". semanage is reasonably fast nowadays (except on RHEL6) but rewriting all the logic is a bit too much churn.

ekohl:
@oranenj sure, but that fact should probably be `selinux_python_command` or similar. It should take the platform into consideration. It would replace the clunky `selinux_semanage_is_python3` fact.

oranenj:
I changed the fact. This is technically a breaking change, but the fact was never documented anywhere and was/is an internal implementation detail

PR #303, opened 2019-10-18
Outdated example

ubellavance:

### DIFF
--- a/examples/fcontext.pp
+++ b/examples/fcontext.pp
@@ -1,4 +1,4 @@
 selinux::fcontext{'set-mysql-log-context':
-  seltype => "mysqld_log_t",
-  pathspec => "/u01/log/mysql(/.*)?",
+  seltype => 'mysqld_log_t',
+  pathspec => '/u01/log/mysql(/.*)?',
 }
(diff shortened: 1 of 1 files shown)


examples/fcontext.pp
```diff
@@ -1,4 +1,4 @@
 selinux::fcontext{'set-mysql-log-context':
-  context  => 'mysqld_log_t',
-  pathname => '/var/tmp/mysql',
+  seltype => "mysqld_log_t",
```
ekohl: Single quotes is still the right way. Could you fix this?

ubellavance: I just took what's at https://github.com/voxpupuli/puppet-selinux/blob/master/manifests/fcontext.pp#L5 so that anyone can use this example with the master branch and not fail. Would we have to change the example in the manifest as well?

ekohl: Ideally yes. Currently the build is failing on linting but we can't lint examples. That would actually be a nice idea :)

ubellavance: I made the change for the example.

PR #306, opened 2019-10-19
Regenerate REFERENCE.md

bastelfreak:

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -12,7 +12,7 @@ _Public Classes_
 _Private Classes_
 
 * `selinux::build`: Configure the system for module building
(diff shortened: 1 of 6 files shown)


REFERENCE.md
```diff
@@ -284,8 +284,8 @@ selinux::fcontext::equivalence
 
 ```puppet
 selinux::fcontext{'set-mysql-log-context':
-  seltype => "mysqld_log_t",
-  pathspec => "/u01/log/mysql(/.*)?",
+  seltype => 'mysqld_log_t',
```
ekohl: Maybe the example should also be aligned?

PR #307, opened 2019-10-20
update example to match current version

ubellavance:


oranenj:
This seems trivial enough that I'll just merge it outright; it can't be more broken than the current example :-P

Thanks!

PR #310, opened 2019-10-21
add partial support for Debian 10, and supporting changes

### LINKED ISSUE #309: No autorelabel when enabling SELinux
This is for documentation purposes, I have a PR for it.

- Puppet: 6.10.1
- Ruby: 2.5.7
- Distribution: Debian 10
- Module version: 3.0.0

On a system with SELinux disabled, create a new file covered by a fcontext rule (for example, `/root/test`).

Apply :
```puppet
class { 'selinux': mode => 'permissive' }
```

Reboot and `ls -Z` the file.

## What are you seeing

The file is `unlabeled_t`.

The file should have a type, such as `user_home_t`.

The `/.autorelabel` file this module creates should be present, but empty. Otherwise the call to `fixfiles` on boot fails because of invalid arguments.

tequeter:

This PR adds support for Debian 10 (and possibly variants like Ubuntu).

The defined types that rely on the SELinux Ruby gem do not work, as it is [not currently shipped](https://tickets.puppetlabs.com/browse/PA-2985) with the Puppet Agent package for Debian.

This PR grew to include 3 refactoring commits and 1 bugfix to support the new feature. Sorry for the overall size.

Fixes #309
Obsoletes and closes #300


manifests/config.pp
```diff
@@ -26,6 +22,16 @@
   }
 
   if $_real_mode {
+    if $facts['os']['family'] == 'Debian' and !$facts['selinux'] {
+      # Debian-based OSes also need to change the kernel boot parameters in the
+      # appropriate version of GRUB.
+      # See: https://wiki.debian.org/SELinux/Setup.
+      exec { 'activate-selinux':
```
bastelfreak: I'm not sure if this is the best approach. Puppet will execute this resource until the node got rebooted.

manifests/config.pp
```diff
@@ -55,7 +61,10 @@
         ensure  => 'file',
         owner   => 'root',
         group   => 'root',
-        content => "# created by puppet for disabled to ${_real_mode} switch\n",
+        # The contents of the file are interpreted on most OSes (at least EL7
+        # and Debian 10) as extra options for fixfiles. Anything else causes an
+        # argument error and a failure to relabel.
+        content => '',
```
bastelfreak: we can also drop the `content` parameter.

manifests/init.pp
```diff
@@ -25,32 +29,38 @@
 # @param exec_restorecon Hash of selinux::exec_restorecon resource parameters
 #
 class selinux (
+  # lint:ignore:parameter_order
```
bastelfreak: Why did you add the linter comment? Reordering the parameters is fine.

bastelfreak:
Hi @tequeter, thanks for the awesome work. Can you take a look at the inline comments I made?

manifests/config.pp
tequeter: Yes, this was a trade-off between robustness and idempotency-until-reboot.

I've rewritten the condition to check one of the immediately visible side-effects of the `activate-selinux` script instead.

manifests/config.pp
tequeter: Enforcing emptiness fixes the nodes that have the buggy `/.autorelabel` file from previous versions of the module. The file is not cleared by the boot script, which chokes on the unexpected contents and exits prematurely.

manifests/init.pp
tequeter: I reordered as requested.

I'm not a big fan of that lint check: the arguments should be listed most important first, not pulled-from-Hiera first. It was a nice idea when it was written, though.

tequeter:
Thank you for the review, @bastelfreak ! I've commented above and pushed a new version.

manifests/config.pp
```diff
@@ -26,6 +22,17 @@
   }
 
   if $_real_mode {
+    if $facts['os']['family'] == 'Debian' and !$facts['selinux'] {
+      # Debian-based OSes also need to change the kernel boot parameters in the
+      # appropriate version of GRUB.
+      # See: https://wiki.debian.org/SELinux/Setup.
+      exec { 'activate-selinux':
+        command  => '/usr/sbin/selinux-activate',
+        unless   => shell_join(['/usr/bin/grep', '-q', '^GRUB_CMDLINE_LINUX=.*security=selinux', '/etc/default/grub']),
```
bastelfreak: nice, people rarely use `shell_join()` :(

bastelfreak:
thanks for your work!

PR #313, opened 2019-12-06
modulesync 2.12.0

dhoppe:


dhoppe:
We need to get rid of the anchor resources.

ekohl:
I've rebased this and disabled the anchor resource check because the way it's used here can be considered part of the API and is complicated. I'd prefer to get modulesync up to date and later look at removing if desired.

ekohl:
:green_heart: now

PR #314, opened 2019-12-20
Amazon Linux v2 should use the new fcontext filetype format

yitsushi:

Amazon Linux v2 is part of the RedHat family, but they started to use their own versioning. Because of that, the originally introduces check for older RHEL distributions thinks Amazon Linux v2 is an older version (`operatingsystemmajrelease == 2`) and tries to use `all files` instead of `a` which is not working.

n/a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,9 @@ group :test do
   gem 'coveralls',                                                  :require => false
   gem 'simplecov-console',                                          :require => false
   gem 'parallel_tests',                                             :require => false
+  # we require unreleased Fedora 30 fact sets
+  # https://github.com/camptocamp/facterdb/commit/1bc038fd42aac53344ac1ee9c129cbccf5b5c0aa
+  gem 'facterdb',                                                   :require => false, :git => 'https://github.com/camptocamp/facterdb.git', :branch => 'master'
 end
 
 group :development do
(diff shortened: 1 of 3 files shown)


bastelfreak:
Hi @yitsushi, thanks for the PR. Can you take a look at the failing spec tests? Can you add amazon linux to the metadata.json? And it would be really awesome if you could have a look at https://github.com/camptocamp/facterdb and provide amazon fact sets (not required to get this merged, but it would help us alot).

yitsushi:
I will try to check what went wrong with spec-test, but currently on my system:
```
2100 examples, 0 failures

[I] ❯ ruby --version
ruby 2.4.4p296 (2018-03-28 revision 63013) [x86_64-darwin18]
```

bastelfreak:
this currently fails because a required change in the facterdb isn't released yet. Please apply the following patch:
```diff
diff --git a/Gemfile b/Gemfile
index f0381cf..b56cf32 100644
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,9 @@ group :test do
   gem 'coveralls',                                                  :require => false
   gem 'simplecov-console',                                          :require => false
   gem 'parallel_tests',                                             :require => false
+  # we require unreleased Fedora 30 fact sets
+  # https://github.com/camptocamp/facterdb/commit/1bc038fd42aac53344ac1ee9c129cbccf5b5c0aa
+  gem 'facterdb',                                                   :require => false, :git => 'https://github.com/camptocamp/facterdb.git', :branch => 'master'
 end
 
 group :development do
```

bastelfreak:
Thanks for the PR!

PR #315, opened 2019-12-20
Update example

ghoneycutt:
Fixes #303


ghoneycutt:
Just took the old PR #303 and fixed the style.

ekohl:
I wonder how the tests are broken by this. Is master already broken?

ghoneycutt:
@ekohl Looks like. Could you approve this one as it passes lint and as an example would not have an effect on spec tests?

ekohl:
Makes sense. Thanks!

PR #317, opened 2020-03-06
WIP: Add selinux_clear_context_cache type

alexjfisher:


ekohl:
Doesn't appear to work:
```text
centos7-64.example.com 20:28:34$ puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest.pp.VmGuLW
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Notice: Compiled catalog for centos7-64.example.com in environment production in 1.40 seconds
  Info: Applying configuration version '1583522917'
  Notice: /Stage[main]/Pulpcore::Install/Package[python3-pulpcore]/ensure: created
  Notice: /Stage[main]/Pulpcore::Install/Package[pulpcore-selinux]/ensure: created
  Info: /Stage[main]/Pulpcore::Install/Package[pulpcore-selinux]: Scheduling refresh of Selinux_clear_context_cache[clear the selinux cache after pulpcore-selinux foo]
  Notice: /Stage[main]/Pulpcore::Install/Selinux_clear_context_cache[clear the selinux cache after pulpcore-selinux foo]: Triggered 'refresh' from 1 event
  # ... LOTS OF OUTPUT HERE ...
  Notice: /Stage[main]/Pulpcore::Config/File[/var/lib/pulp/tmp]/ensure: created
  # ... LOTS OF OUTPUT HERE ...
  Notice: Applied catalog in 120.59 seconds

centos7-64.example.com executed in 126.52 seconds
Exited: 2
    applies with no errors

centos7-64.example.com 20:30:41$ mktemp -t apply_manifest.pp.XXXXXX
  /tmp/apply_manifest.pp.zhQ0Qm

centos7-64.example.com executed in 0.04 seconds
localhost $ scp /tmp/beaker20200306-535126-eof8hc centos7-64.example.com:/tmp/apply_manifest.pp.zhQ0Qm {:ignore => }

centos7-64.example.com 20:30:41$ puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest.pp.zhQ0Qm
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Info: Loading facts
  Notice: Compiled catalog for centos7-64.example.com in environment production in 1.48 seconds
  Info: Applying configuration version '1583523044'
  Notice: /Stage[main]/Pulpcore::Config/File[/var/lib/pulp/tmp]/seltype: seltype changed 'var_lib_t' to 'pulpcore_var_lib_t'
  # ... LOTS MORE OUTPUT with refreshes ...
```

oranenj:
This seems like a bit of a hack, but I'm not opposed to the idea.

Reading the manpage of the matchpathcon functions, I don't see any immediate safety concerns either. As far as I understand, matchpathcon and matchpathcon_init will just reload things into memory as needed if they've been reset by matchpathcon_fini.

oranenj:
@ekohl do you know if Puppet uses multiple threads when applying manifests? Looking at the matchpathcon source code, it seems to use thread-local storage in some manner that might affect this.
Also, at what point are the file resource's initial attributes loaded? If it happens before the reset resource is refreshed, then it will have already fetched the wrong context for the file.

ekohl:
We had a discussion on IRC about this. It may be that the catalog is caching the defaults via
https://github.com/puppetlabs/puppet/blob/d8528c87ad067399fe47cb4972b913f588484679/lib/puppet/type/file/selcontext.rb#L126
That would then become part of the catalog and can't be changed later.

My reproducer is https://github.com/theforeman/puppet-pulpcore/blob/a0796b669db347bc00da1b4a395d2d2ebf75e303/spec/acceptance/basic_spec.rb#L6-L21.

alexjfisher:
That's a real pain, but not necessarily a dead-end.

Looks like Puppet would need a bit of modifying. Basically change the defaults to a place holder `:default_context`.  Then figure out the right combination of `should` `sync`, `insync?` etc. that would need overriding/tweaking.
I got as far as https://gist.github.com/alexjfisher/5656355aa542e4366c70a3091b1cdfcc

PR #324, opened 2020-07-24
modulesync 3.0.0 + code reformatting because of puppet-lint updates

bastelfreak:

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -11,7 +11,6 @@ class selinux::config (
   $mode,
   $type,
 ) {
-
   assert_private()
 
   if ($mode == 'enforcing' and !$facts['os']['selinux']['enabled']) {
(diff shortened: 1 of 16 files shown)


manifests/config.pp
```diff
@@ -57,7 +56,7 @@
     # a complete relabeling is required when switching from disabled to
     # permissive or enforcing. Ensure the autorelabel trigger file is created.
     if $_real_mode in ['enforcing','permissive'] and
-      !$facts['os']['selinux']['enabled'] {
+    !$facts['os']['selinux']['enabled'] {
```
ekohl, line 59: This looks ugly. Perhaps it needs braces so the lint plugin understands it?

bastelfreak: fixed it

PR #325, opened 2020-07-25
modulesync 3.0.0 & puppet-lint updates

bastelfreak:
modulesync 3.0.0


ekohl:
Looks like #324 already took care of this.

PR #330, opened 2020-10-14
semanage_ports.py: fix shebang

kibidouil:
ERROR: ambiguous python shebang in /var/lib/puppet/default/modules/selinux/lib/puppet_x/voxpupuli/selinux/semanage_ports.py: #!/usr/bin/python. Change it to python3 (or python2) explicitly.
Don't work with python2, work with python3.

### DIFF
--- a/data/os/RedHat/CentOS/7.yaml
+++ b/data/os/RedHat/CentOS/7.yaml
@@ -1,4 +1,2 @@
 ---
-selinux::package_name:
-  - policycoreutils-python
-  - python36
+selinux::package_name: policycoreutils-python
--- a/data/os/RedHat/CentOS/7.yaml
+++ b/data/os/RedHat/CentOS/7.yaml
@@ -1,2 +1,4 @@
 ---
-selinux::package_name: policycoreutils-python
+selinux::package_name:
+  - policycoreutils-python
+  - python36
--- a/data/os/RedHat/CentOS/8.yaml
+++ b/data/os/RedHat/CentOS/8.yaml
@@ -1,0 +1,2 @@
+---
+selinux::package_name: policycoreutils-python
--- a/data/os/RedHat/RedHat/7.yaml
+++ b/data/os/RedHat/RedHat/7.yaml
@@ -1,4 +1,2 @@
 ---
-selinux::package_name:
-  - policycoreutils-python
-  - python36
+selinux::package_name: policycoreutils-python
--- a/data/os/RedHat/RedHat/7.yaml
+++ b/data/os/RedHat/RedHat/7.yaml
@@ -1,2 +1,4 @@
 ---
-selinux::package_name: policycoreutils-python
+selinux::package_name:
+  - policycoreutils-python
+  - python36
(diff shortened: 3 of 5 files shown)


kenyon:
It would be good if we could also have tests which currently fail on CentOS 8, and then pass with this patch applied. Otherwise LGTM.

kibidouil:
> It would be good if we could also have tests which currently fail on CentOS 8, and then pass with this patch applied.

I'm sorry but I can't do that. I would change the shebang because I want to make a rpm wich include the module. Now when you build a rpm with CentOS8, there is a script (/usr/lib/rpm/redhat/brp-mangle-shebangs) checking shebang for executable scripts.
The module working fine, but can't be added to a rpm. That's not really an issue...
 
> Shouldn't we use the shebang line that the Python docs recommend, `#!/usr/bin/env python3` ?
Yes, we should.

kenyon:
OK, there aren't any tests for this file currently, so this looks good to me.

kenyon (approved):

ekohl (commented):
I think this breaks EL7 which is still supported. Because it's SELinux, you really need to run it on real VMs since Docker can't work.

kibidouil:
> I think this breaks EL7 which is still supported. Because it's SELinux, you really need to run it on real VMs since Docker can't work.

I don't think so, EL7 supports Python3 since 7.7 release.

ekohl:
Supports doesn't mean it's present. AFAIK there's nothing in this module that ensures it's present. I don't even know if the SELinux library is present for EL7.

kibidouil:
> Supports doesn't mean it's present. AFAIK there's nothing in this module that ensures it's present. I don't even know if the SELinux library is present for EL7.

Ok, now, I think it's present

ekohl:
>  Don't work with python2, work with python3.

I still think this shouldn't be true and you should be able to use Python 2 on EL7. Can you explain why you need to pull in Python 3?

kibidouil:
> > Don't work with python2, work with python3.
> 
> I still think this shouldn't be true and you should be able to use Python 2 on EL7. Can you explain why you need to pull in Python 3?

I just want to use python3 because python2 is deprecated CentOS 8. In Centos/RH 8, /usr/bin/python doesn't exist anymore. you must choose python3 or 2. When I try python 2, the script doesn't work.

ekohl:
That's why we have the [selinux_python_command fact](https://github.com/voxpupuli/puppet-selinux/blob/master/lib/facter/selinux_python_command.rb). And here is where we hook it up:
https://github.com/voxpupuli/puppet-selinux/blob/1f5bb7f045c103fbf25c4410264b5b7b20fb9851/lib/puppet/provider/selinux_port/semanage.rb#L8-L13

It doesn't use the shebang:
https://github.com/voxpupuli/puppet-selinux/blob/1f5bb7f045c103fbf25c4410264b5b7b20fb9851/lib/puppet/provider/selinux_port/semanage.rb#L55

So I don't really see the point of this PR and what it's solving. I thought the problem was already solved.

kibidouil:
So, the shebang is not needed at all. It is possible to remove it or to set the file not executable ?
When you want to include this module in a rpm, the build check executables files and test the shebang.

ekohl:
I think that should be possible.

kibidouil:
So, I remove the executable bit and python36 for EL7. I've restore the old shebang.

PR #334, opened 2020-11-25
modulesync 4.0.0

bastelfreak:
modulesync 4.0.0


ekohl (changes requested):
This is an exception: there are acceptance tests but they can't run under Docker. I think we should manually modify the actions file for now.

ekohl (approved):

PR #336, opened 2021-02-14
Deprecate selinux_python_command fact and move handling into provider

### LINKED ISSUE #335: The semanage_ports.py script needs to pick the correct version of python on a system with multiple installed

- Puppet: Any
- Ruby: Any
- Distribution: EL8
- Module version: Any

## Information

On systems with multiple python interpreters installed, one needs to be explicitly chosen so that the correct libraries are used.

I'm not exactly sure what the correct method for doing this is but it may be better to detect the system python using Ruby and then generate the script on the fly with the correct interpreter in place.

### LINKED ISSUE #342: Add support for Puppet 7
The module works properly with Puppet 7 and should be updated to reflect this.

### LINKED ISSUE #343: Add support for Fedora 33
The module functions properly under Fedora 33 and should be updated to reflect this.

trevor-vaughan:

Fixed:
  * Moved the python discovery into the semanage_ports provider
  * Ensure that the correct python is chosen by attempting to load the
    `semanage` library
  * Removed the #! line from the semanage_ports.py file to fix RPM
    building
  * Removed CentOS 6 support from `test-acceptance-with-vagrant` since
    it will no longer function due to upstream deprecation

Changed:
  * Marked the `selinux_python_command` fact as deprecated in the README
    since the logic now resides in the provider itself
  * Added support for Puppet 7
  * Added Fedora 33 support

Fixes #335
Fixes #342 
Fixed #343

### DIFF
--- a/lib/puppet/provider/selinux_port/semanage.rb
+++ b/lib/puppet/provider/selinux_port/semanage.rb
@@ -5,7 +5,37 @@ Puppet::Type.type(:selinux_port).provide(:semanage) do
   # SELinux must be enabled. Is there a way to get a better error message?
   confine selinux: true
 
-  python_command = Facter.value(:selinux_python_command)
+  # Determine the appropriate python command
+  def self.python_command
+    @python_command ||= nil
+    return @python_command unless @python_command.nil?
+
+    # Find the correct version of python on the system
+    python_paths = [
+      '/usr/libexec/platform-python',
+      'python',
+      'python3',
+      'python2'
+    ]
+
+    python_command = nil
+    python_paths.each do |pypath|
+      candidate = Puppet::Util.which(pypath)
+
+      next unless candidate
+      if Puppet::Util::Execution.execute("#{candidate} -c 'import semanage'", failonfail: false).exitstatus.zero?
+        python_command = candidate
+        break
+      end
+    end
+
+    @python_command = python_command
+
+    @python_command
+  end
+
+  confine false: python_command.nil?
+
   # current file path is lib/puppet/provider/selinux_port/semanage.rb
   # semanage_ports.py is lib/puppet_x/voxpupuli/selinux/semanage_ports.py
   PORTS_HELPER = File.expand_path('../../../../puppet_x/voxpupuli/selinux/semanage_ports.py', __FILE__)
--- a/metadata.json
+++ b/metadata.json
@@ -54,7 +54,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 7.0.0"
+      "version_requirement": ">= 5.5.8 < 8.0.0"
     }
   ]
 }
--- a/metadata.json
+++ b/metadata.json
@@ -29,7 +29,8 @@
       "operatingsystemrelease": [
         "30",
         "31",
-        "32"
+        "32",
+        "33"
       ]
     },
     {
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ running system.
 
 ## Known problems / limitations
 
+* The `selinux_python_command` fact is now deprecated and will be removed in
+  version 4 of the module.
 * If SELinux is disabled and you want to switch to permissive or enforcing you
   are required to reboot the system (limitation of SELinux). The module won't
   do this for you.
--- a/lib/facter/selinux_python_command.rb
+++ b/lib/facter/selinux_python_command.rb
@@ -1,3 +1,4 @@
+# DEPRECATED: Determine the path to python on the system
 Facter.add(:selinux_python_command) do
   confine osfamily: 'RedHat'
   setcode do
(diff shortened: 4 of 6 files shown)


lib/puppet/provider/selinux_port/semanage.rb
```diff
@@ -5,7 +5,26 @@
...
+  python_paths.each do |pypath|
+    candidate = Puppet::Util.which(pypath)
+
+    next unless candidate
+    if Puppet::Util::Execution.execute("#{candidate} -c 'import semanage'", failonfail: false).exitstatus.zero?
+      python_command = candidate
+      break
+    end
+  end
+  confine false: python_command.nil?
```
ekohl, line 26: I think this implementation means it also runs this code if the provider is not used. Is it possible to pass a proc to confine?

An alternative is to ship a wrapper (ba)sh script that calls exec at the end but I'm not sure if you can get the path right to a pluginsynced script.

trevor-vaughan: @ekohl Yeah, you can put a proc in a confine. I guess I can drop in an instance variable and see if I can set it in the proc.

trevor-vaughan: To be fair, the old fact ran every time as well so this isn't worse (it's just not better)

ekohl: I also had a look at `optional_command` but I'm not sure if that's better. Btw, how about this:

```ruby
def find_python_command
  python_paths.detect do |pypath|
    Puppet::Util::Execution.execute("#{pypath} -c 'import semanage'", failonfail: false).exitstatus.zero?
  end
end
```

I don't think you need to find an absolute path, the command should fail otherwise anyway, right?

trevor-vaughan:
@ekohl Reworked the discovery part and it seems to function properly and should only trigger when the confine is processed.

lib/puppet/provider/selinux_port/semanage.rb
```diff
@@ -5,7 +5,37 @@
...
+    python_command = nil
+    python_paths.each do |pypath|
+      candidate = Puppet::Util.which(pypath)
+
+      next unless candidate
+      if Puppet::Util::Execution.execute("#{candidate} -c 'import semanage'", failonfail: false).exitstatus.zero?
+        python_command = candidate
+        break
+      end
+    end
```
ekohl, line 30: Would this work?

```suggestion
    python_command = python_paths.detect do |pypath|
      Puppet::Util::Execution.execute("#{pypath} -c 'import semanage'", failonfail: false).exitstatus.zero?
    end
```

metadata.json
```diff
@@ -29,7 +29,8 @@
       "operatingsystemrelease": [
         "30",
         "31",
-        "32"
+        "32",
```
ekohl, line 32: I don't mind these changes, but could you add them in a separate PR for a complete changelog?

lib/puppet/provider/selinux_port/semanage.rb
trevor-vaughan: @ekohl I don't think it will.

`failonfail` doesn't do what you think if the command itself doesn't exist so we need the `next unless candidate` bit.

trevor-vaughan:
@ekohl Updated to individual commits.

trevor-vaughan:
@ekohl Are you OK with the explanation above?

oranenj:
It might actually make more sense to get rid of the python helper for semanage port listing entirely, nowadays.

I had to write it for selinux 1.0.0 because back then, semanage was *unbeliveably* slow. Nowadays, on any non-ancient RHEL (7 and 8), semanage is fast enough that the python script should no longer be necessary.

ekohl:
If we can get rid of it altogether, that would be nicer. @trevor-vaughan any thoughts on that?

trevor-vaughan:
@ekohl Sorry, I lost track of this. I'll try to see if I can flip it this weekend.

trevor-vaughan:
@oranenj and @ekohl I tinkered around with things and I think it would be best to leave the script for now.

Running `semanage_ports.py`:

```
real    0m0.082s
user    0m0.068s
sys     0m0.013s
```

Running `semanage port -l`:

```
real    0m1.571s
user    0m1.480s
sys     0m0.089s
```

Why this is the case, I have no idea, but we have something that works and is quite a bit faster than the underlying command so I'd like to stick with it for now if possible.

ekohl:
> * Added support for Puppet 7
> 
> * Added Fedora 33 support

Could you at least file issues for this and make this PR resolve them so the changelog is correct?

trevor-vaughan:
@ekohl Updated and referenced above

ekohl (approved):
I've taken the liberty to change the PR title to be more descriptive for the changelog. Thanks!

PR #338, opened 2021-02-26
Make package class public

### LINKED ISSUE #337: Error finding provider when using selinux_port
I would like to use aelinux_port and other types/and providers provided by this module but not use the whole module to control selinux.

This requires that the packages for selinux be installed.  This module already has the logic for that in selinux::packages and all the hiera data that is set up but I can't use it without the whole module because it it private.

I have a fix that makes the package module public so it can be included in other modules when the type or providers are used  but does not require them to include the whole module.  It also will not conflict with the module if they do choose to include the module to control selinux.

jeannegreulich:
- Update the selinux::package class to be public.
  The packages needed to support the various types and providers from this
  module vary between OS families and versions.  Making the this
  class available to users who wish to use the types but not use the
  entire module will prevent them from having the recreate the logic of determining
  what package needs to be installed.

 
Fixes #337

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+- Change the selinux::package module to be public.  The packages needed to support the various
+  types and providers from this module vary between OS families and versions.  Making the this
+  class available to users who wish to use the types but not use the entire module will prevent
+  them from having te recreate the logic of determing what package needs to be installed.
+
 ## [v3.2.0](https://github.com/voxpupuli/puppet-selinux/tree/v3.2.0) (2020-04-30)
 
 [Full Changelog](https://github.com/voxpupuli/puppet-selinux/compare/v3.1.0...v3.2.0)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+- Change the selinux::package module to be public.  The packages needed to support the various
+  types and providers from this module vary between OS families and versions.  Making the this
+  class available to users who wish to use the types but not use the entire module will prevent
+  them from having te recreate the logic of determing what package needs to be installed.
+
```
kenyon, line 11: The changelog will be autogenerated when the next release is prepared, so no need to add this.

kenyon:
Is this really necessary? From the readme:

> Basic usage
> `include selinux`
> This will include the module and allow you to use the provided defined types, but will not modify existing SELinux settings on the system.

jeannegreulich:
Some of the types created by this module are useful in other modules but don't want those modules include the entire selinux  module.  It is necessary that the packages be installed for the types like selinux_port.  One could put a a package resource in with the selinux_port  but because the package names differ so widely from version to version it is handy to use the data already included in this module and just include selinux::package.

jeannegreulich:
Yep your right. Sorry.

PR #339, opened 2021-03-24
Support Puppet 7

mattock:

This allows using Puppet 7 with this module when strict dependency checking is enabled.

Without this change Kafo installers such as Foreman's Kafo installer will fail with this error message:

```
Puppet 7.5.0 does not meet requirements for puppet-selinux (>= 5.5.8 < 7.0.0)
Cannot continue due to incompatible version of Puppet.
Use --skip-puppet-version-check to disable this check.
```

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -54,7 +54,7 @@
   "requirements": [
     {
       "name": "puppet",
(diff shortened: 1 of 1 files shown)


ekohl (approved):
Let's see if the tests pass :)

PR #346, opened 2021-05-19
Add package autorequires to native types

### LINKED ISSUE #345: Resources that use 'semanage' should have an autorequires on the various packages
After using the latest release of the module, I noticed that `selinux_port` was having issues in some of my tests.

This turned out to be that (for whatever reason) the package containing the python `semanage` library had not yet been installed.

Instead of requiring users to include the entire module, I would like to propose that all commands that need `semanage` have an `autorequires` added on the various potential package names.

Unfortunately, this means that a failure in a package update will cascade but this is pretty standard and will fix potential 'dual-run' issues moving forward.

The main drawback is that, as new OS flavors are added, the package list will need to be updated to reflect things. I don't think that requiring the `selinux::package` class is a good idea because it contains more than the single package that is required for obtaining the `semanage` command.

trevor-vaughan:

Changed:

  * Added `autorequire` calls to ensure that the necessary packages are
    installed if `semanage` is to be used.

Closes #345


ekohl (changes requested):
On Fedora it's now `python3-policycoreutils` which provides the actual library we use. Please add that too. Otherwise I'm happy to see this. Overall I'm a big fan of `autorequire`.

trevor-vaughan:
@ekohl Done!

ekohl (approved):

trevor-vaughan:
Hmm...for some reason, this doesn't seem to be resolving the issue, investigating.

ekohl:
autorequire only creates `Package[$package] -> MyResource[$x]` relationship but you are responsible for `ensure_packages([$package])`. Are you sure it's in your catalog?

trevor-vaughan:
:facepalm: Needs to be on `policycoreutils-python-utils`. Missed that one somehow.

trevor-vaughan:
I'm starting to doubt that `confine` statements happen when a provider is first called :-(

trevor-vaughan:
Looks like I need to add a feature...neat

trevor-vaughan:
Since this is used in `prefetch` I'm actually not sure if anything is going to work to make it one-run clean and I have no idea why nobody has hit this in the past.

trevor-vaughan:
@ekohl This appears to now work properly as far as practical purposes are concerned

ekohl (approved):
I was wondering about a pattern:
```ruby
def self.python_command
  @python_command ||= begin
    ...
  end
end
```

But i wasn't quite sure if it actually made it cleaner.

trevor-vaughan:
> I was wondering about a pattern:
> 
> ```ruby
> def self.python_command
>   @python_command ||= begin
>     ...
>   end
> end
> ```
> 
> But i wasn't quite sure if it actually made it cleaner.

Well, the alternative is something like:

```
unless @python_command
  @python_command = begin
    ...
  end
end
```

Functionally equivalent and less typing

PR #351, opened 2021-09-17
modulesync 6.0.0

bastelfreak:


bastelfreak:
Fixes:
* https://github.com/voxpupuli/puppet-selinux/issues/357

smortex (approved):
LGTM!

.sync.yml
```diff
@@ -1,10 +1,6 @@
 ---
-.pmtignore:
-  paths:
-    - .puppet-lint.rc
 spec/spec_helper.rb:
   mock_with: ':mocha'
-Gemfile:
-  optional:
-    ':test':
-      - gem: puppet-lint-param-docs
```
ekohl, line 10: I think this line meant we used to check the param docs. Does removal here mean we stop checking that?

lib/puppet/provider/selinux_fcontext/semanage.rb
```diff
@@ -13,9 +15,7 @@
   osversion       = Facter.value('operatingsystemmajrelease')
   operatingsystem = Facter.value('operatingsystem')
   @old_semanage = false
-  if (osfamily == 'RedHat') && (Puppet::Util::Package.versioncmp(osversion, '6') <= 0) && (operatingsystem != 'Amazon')
-    @old_semanage = true
-  end
+  @old_semanage = true if (osfamily == 'RedHat') && (Puppet::Util::Package.versioncmp(osversion, '6') <= 0) && (operatingsystem != 'Amazon')
```
ekohl, line 18: If you're dropping EL6 support, does that mean this is now redundant?

lib/puppet/provider/selinux_login/semanage.rb
```diff
@@ -87,14 +90,14 @@ def self.prefetch(resources)
     instances.each do |provider|
       resource = resources[provider.name]
       if resource
-        unless resource[:selinux_user].to_s == provider.selinux_user && resource[:selinux_login_name].to_s == provider.selinux_login_name || resource.purging?
-          raise Puppet::ResourceError, "Selinux_port['#{resource[:name]}']: title does not match its port and protocol, and a conflicting resource exists"
-        end
+        raise Puppet::ResourceError, "Selinux_port['#{resource[:name]}']: title does not match its port and protocol, and a conflicting resource exists" unless (resource[:selinux_user].to_s == provider.selinux_user && resource[:selinux_login_name].to_s == provider.selinux_login_name) || resource.purging?
```
ekohl, line 93: This really is worse. The problem is that we don't have a cop for the max line length (we've disabled that), which means we end up with these horribly long lines.

manifests/config.pp
```diff
@@ -8,8 +8,8 @@
 # @api private
 #
 class selinux::config (
-  $mode,
-  $type,
+  Optional[Enum['enforcing', 'permissive', 'disabled']] $mode = undef,
+  Optional[Enum['targeted', 'minimum', 'mls']] $type          = undef,
```
ekohl, line 12: Isn't this an API change? Previously it was a required parameter, now it becomes optional.
```suggestion
  Enum['enforcing', 'permissive', 'disabled'] $mode,
  Enum['targeted', 'minimum', 'mls'] $type,
```

.sync.yml
bastelfreak: that's now coming via vox-pupuli-puppet-lint-plugins

lib/puppet/provider/selinux_fcontext/semanage.rb
bastelfreak: you're correct. I cherry-picked and fixed it in https://github.com/voxpupuli/puppet-selinux/pull/368

lib/puppet/provider/selinux_login/semanage.rb
bastelfreak: I agreee. I reverted it

PR #352, opened 2021-10-15
Add Debian 11 support

jbrnfnd:
This PR enables support for Debian 11 by adding a yaml file, 11.yaml in the Debian subfolder
It is a straight renamed copy of 10.yaml

### DIFF
--- a/data/os/Debian/Debian/11.yaml
+++ b/data/os/Debian/Debian/11.yaml
@@ -1,0 +1,6 @@
+---
+selinux::package_name:
+  - policycoreutils-python-utils
+  - selinux-basics
+  - selinux-policy-default
+selinux::refpolicy_package_name: selinux-policy-dev
--- a/metadata.json
+++ b/metadata.json
@@ -36,7 +36,8 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
-        "10"
+        "10",
+        "11"
       ]
     },
     {


ghoneycutt:
Thanks for the contribution! For this module to support Debian 11, it needs to be reocrded in the metadata - https://github.com/voxpupuli/puppet-selinux/blob/master/metadata.json

smortex:
I think you can fix CI by improving:

https://github.com/voxpupuli/puppet-selinux/blob/92f1bf59c43de4e8112b6f9d6f13f36e0a4d3671/manifests/boolean.pp#L39

```diff
-  if $facts['os']['selinux']['enabled'] {
+  if facts('os.selinux.enabled') {
```

jbrnfnd:
Hi,

From the documentation, the squared bracket syntax is also supported in Puppet 5.5
https://puppet.com/docs/puppet/5.5/lang_facts_and_builtin_vars.html#the-factsfact_name-hash
I don't know why this is failing, as the syntax is the same used as in older versions of puppet-selinux
As puppet5 is EOL since end of january 2021, why it is still in the CI workflow?
Wouldn't it be better to only support active Puppet versions?

cmseal:
Hi.  Aside from the EOL Puppet 5 test and Puppet 5 support requested to be removed in a separate PR, is there anything else required to move this forwards?

jbrnfnd:
Hi, due to work related things i am completely stalled here. i would like to post a pull request but don't know how.
the yml files in .github/workflows do not indicate to exclude puppet 5 in my opinion.

ekohl:
It's because of this line:
https://github.com/voxpupuli/puppet-selinux/blob/92f1bf59c43de4e8112b6f9d6f13f36e0a4d3671/metadata.json#L58
If you update that (preferably to `>= 6.1.0`) then it should stop testing on Puppet 5.

laser-dev-support:
Hi, any update on this PR? We really need to have it working on Debian 11.

thanks a lot

f.

ekohl:
I opened https://github.com/voxpupuli/puppet-selinux/pull/361 & https://github.com/voxpupuli/puppet-selinux/pull/362

ekohl:
Solved via https://github.com/voxpupuli/puppet-selinux/pull/359 instead.

PR #356, opened 2022-07-18
add support for selinux login

### LINKED ISSUE #166: selinux::users && selinux::logins
Would you like to implement selinux::users ( semanage user ) and selinux::login ( semanage login )?

bshelton:
This is a PR to add basic support for selinux logins. This will allow mapping of linux users or groups to a selinux user.

Fixes #166

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -91,12 +91,10 @@ class selinux (
   if $exec_restorecon {
     create_resources ( 'selinux::exec_restorecon', $exec_restorecon )
   }
-  if $login {
-    $login.each |$login_name, $login_attributes| {
-      selinux::login { $login_name:
+  $login.each |$login_name, $login_attributes| {
+    selinux::login { $login_name:
       * => $login_attributes,
     }
-}
   }
 
   # Ordering
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -92,7 +92,7 @@ class selinux (
     create_resources ( 'selinux::exec_restorecon', $exec_restorecon )
   }
   if $login {
-    $login.each |$login_name, $login_attributes {
+    $login.each |$login_name, $login_attributes| {
       selinux::login { $login_name:
       * => $login_attributes,
     }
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -89,6 +91,9 @@
   if $exec_restorecon {
     create_resources ( 'selinux::exec_restorecon', $exec_restorecon )
   }
+  if $login {
```
bastelfreak, line 94: while this style is consistent with the rest of the class, it doesn't meet best practices. using create_resources is deprecated. Instead I sugges to default $login to an empty hash and iterate on it

```puppet
$login.each |$login_name, $login_attributes {
  selinux::login { $login_name:
    * => $login_attributes,
  }
}
```

manifests/login.pp
```diff
@@ -0,0 +1,44 @@
...
+#     selinux_login_name  => 'localuser',
+#     selinux_user => 'staff_u',
+#   }
+#
+# @param ensure Set to present to add or absent to remove a selinux login.
+# @param selinux_login_name A Linux user or group
+# @param selinux_user The selinux user to map to
+#
+define selinux::login (
+  String                    $selinux_login_name,
```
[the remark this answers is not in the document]
bshelton: Yep.. sorry I'm lacking in experience with Ruby.

manifests/init.pp
```diff
@@ -54,6 +55,7 @@
   Optional[Hash] $permissive      = undef,
   Optional[Hash] $port            = undef,
   Optional[Hash] $exec_restorecon = undef,
+  Optional[Hash] $login           = undef,
```
bastelfreak, line 58: please switch this to:
```suggestion
  Hash[String[1],Hash[String[1],String[1]]] $login = {},
```

This should allow an empty hash or an hash with a string as key and another hash as value

manifests/init.pp
bastelfreak: you can drop the `if $login`. `$login.each` will notice that $login is empty and not iterate on it.

bastelfreak:
thanks for the work!

Issue #357, opened 2022-07-23
Drop Puppet 5 support; require 6.1. or newer


bastelfreak:
Puppet 5 is EoL and it won't pass our CI anymore

bastelfreak:
implemented in #365

PR #359, opened 2022-09-22
Add Debian 11 support by moving the policy packages definition

jcpunk:
Provide default package names for Debian based systems.  Can still override for specific versions if need be.

One of my internal modules tests against Debian 11.  I didn't guard the selinux bits and noticed the package names don't have a default set for Debian.

### DIFF
--- a/data/os/Debian.yaml
+++ b/data/os/Debian.yaml
@@ -1,4 +1,9 @@
 ---
+selinux::package_name:
+  - policycoreutils-python-utils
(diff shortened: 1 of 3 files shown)


smortex:
What about adding Debian 11 to metadata.json and update this PR to something like "Add support for Debian 11"?

smortex (approved):
LGTM! Puppet 5 being EOL, we can drop support for it (#357 ; #351)

PR #360, opened 2022-09-22
Add param for selinux::fcontext::equivalence elements

### LINKED ISSUE #353: Using selinux::fcontext::equivalence with hiera
Hello

I'm attempting to use selinux module 3.4.1 with hiera and am trying to avoid having to write a class for a particular node where i need to use a fcontext equivalence. 

I can easily set boolean values via including the class and creating a hiera hash, and I can setup fcontexts.  However - I was surprised that defining a hash for selinux::fcontext::equivalence had no effect. 

I see there's not 'create_resources' statement in init.pp for equivalences as there are for the other defines. I'm sure this by design, rather than a bug bot if so is there a way to create equivalences via hiera directly?

Thanks

jcpunk:
Class parameter for selinux::fcontext::equivalence elements

Fixes #353

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -27,6 +27,7 @@
 # @param default_builder which builder to use by default with selinux::module
 # @param boolean Hash of selinux::boolean resource parameters
 # @param fcontext Hash of selinux::fcontext resource parameters
+# @param fcontext_equivalence Hash of selinux::fcontext::equivalence resource parameters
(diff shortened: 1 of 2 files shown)


ekohl:
Looks like the test failures are related.

jcpunk:
I think I've got it sorted now.

kenyon (approved):

ekohl (approved):

PR #361, opened 2022-12-09
Drop Puppet 5 support

ekohl:

PR #363, opened 2023-03-09
Wrap some more commands in "if selinux is enabled"

jcpunk:
Some selinux commands only work when selinux is enabled.  There are a few that were missed in previous updates.


jcpunk:
Github isn't displaying the module.pp diff well, it really is just adding an `if` wrapper.

kenyon (commented):
Maybe do #361 first to get rid of the failing Puppet 5 test?

kenyon (approved):

PR #368, opened 2023-06-29
Drop EoL RHEL 6 support

bastelfreak:
cherry-picked from https://github.com/voxpupuli/puppet-selinux/pull/351

### DIFF
--- a/lib/puppet/provider/selinux_fcontext/semanage.rb
+++ b/lib/puppet/provider/selinux_fcontext/semanage.rb
@@ -9,13 +9,7 @@ Puppet::Type.type(:selinux_fcontext).provide(:semanage) do
 
   mk_resource_methods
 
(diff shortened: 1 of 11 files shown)


lib/puppet/provider/selinux_fcontext/semanage.rb
```diff
@@ -9,13 +9,7 @@
 
   mk_resource_methods
 
-  osfamily        = Facter.value('osfamily')
-  osversion       = Facter.value('operatingsystemmajrelease')
-  operatingsystem = Facter.value('operatingsystem')
   @old_semanage = false
```
ekohl, line 12: This is now redundant. So you can also simplify it further: https://github.com/voxpupuli/puppet-selinux/pull/372

PR #375, opened 2023-08-21
selinux_build_module_simple.sh: improve quoting

kenyon:
If module_name or module_dir had a space, this script would fail.

Also avoid existence test for the tmp dir and use mkdir -p instead.

### DIFF
--- a/files/selinux_build_module_simple.sh
+++ b/files/selinux_build_module_simple.sh
@@ -4,13 +4,13 @@ module_dir="$2"
 
 set -e
 
-cd $module_dir
-test -d tmp || mkdir tmp
+cd "$module_dir"
+mkdir -p tmp
 
-checkmodule -M -m -o tmp/${module_name}.mod ${module_name}.te
-package_args="-o ${module_name}.pp -m tmp/${module_name}.mod"
+checkmodule -M -m -o "tmp/${module_name}.mod" "${module_name}.te"
+
 if [ -s "${module_name}.fc" ]; then
-    package_args="${package_args} --fc ${module_name}.fc"
+    semodule_package -o "${module_name}.pp" -m "tmp/${module_name}.mod" --fc "${module_name}.fc"
+else
+    semodule_package -o "${module_name}.pp" -m "tmp/${module_name}.mod"
 fi
-
-semodule_package ${package_args}
--- a/manifests/module.pp
+++ b/manifests/module.pp
@@ -136,7 +136,7 @@ define selinux::module (
       exec { "build-module-${title}":
         path    => '/bin:/usr/bin',
         cwd     => $module_dir,
-        command => "${build_command} || (rm -f ${module_file}.pp ${module_file}.loaded && exit 1)",
+        command => "${build_command} || (rm -f '${module_file}.pp' '${module_file}.loaded' && exit 1)",
         creates => "${module_file}.pp",
         notify  => Exec["install-module-${title}"],
       }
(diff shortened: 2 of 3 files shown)


files/selinux_build_module_simple.sh
```diff
@@ -4,13 +4,13 @@ module_dir="$2"
...
 
-checkmodule -M -m -o tmp/${module_name}.mod ${module_name}.te
+checkmodule -M -m -o "tmp/${module_name}.mod" "${module_name}.te"
 package_args="-o ${module_name}.pp -m tmp/${module_name}.mod"
 if [ -s "${module_name}.fc" ]; then
     package_args="${package_args} --fc ${module_name}.fc"
 fi
 
-semodule_package ${package_args}
+semodule_package "${package_args}"
```
smortex, line 16: This change behavior.  Assuming we do not enter the if branch:

#### Before

`semodule_package` got 4+ arguments:
1. `-o`
2. `${module_name}.pp`
3. `-m`
4. `tmp/${module_name}.mod`

With the issue that when `${module_name}` has spaces, the parameter 2 and 4 are split and this break.

#### Now

`semodule_package` get 1 arguments:
1. `-o ${module_name}.pp -m tmp/${module_name}.mod`

This does not feel right.

#### Proposal

Spiting each parameter in a variable might help:

```sh
# crappy name for the example
o="${module_name}.pp"
m="tmp/${module_name}.mod"

#[...]

if [ -s "${module_name}.fc" ]; then
  semodule_package -o "${o}" -m "${m}" --fc "${module_name}.fc"
else
  semodule_package -o "${o}" -m "${m}"
end
```

kenyon: Ah yeah I shouldn't have quoted that line. Only the `checkmodule` line needs quoting. Shell scripts considered harmful 😉

kenyon: Actually the `semodule_package` does need to be split also, let me do that.

kenyon: Fixed.

smortex (approved):
That look more correct!

I guess the CI don't the case with spaces but maybe you have a real-case scenario you can check against this branch to ensure correctness?

LGTM!

kenyon:
Yeah, that's another problem with using a shell script like this, it would have to be tested in acceptance tests.

I found more command lines where filenames should be quoted. Tomorrow I'll see if I can add an acceptance test that fails without this PR.

smortex:
Thinking about it, the "best" is probably to rely on [shellquote](https://www.puppet.com/docs/puppet/7/function.html#shellquote) to escape shell special chars.  I can't really think that somebody able to inject malicious subcommands in a control-repo without validation has no prior root access to the system, but for the sake of best practice it may make sense to prevent such misuse.

kenyon:
> Thinking about it, the "best" is probably to rely on [shellquote](https://www.puppet.com/docs/puppet/7/function.html#shellquote) to escape shell special chars. I can't really think that somebody able to inject malicious subcommands in a control-repo without validation has no prior root access to the system, but for the sake of best practice it may make sense to prevent such misuse.

Hmmmm, the command line is actually already shellquoted: https://github.com/voxpupuli/puppet-selinux/blob/37e1761d6df0a02439416bcc17e384a84249465d/manifests/module.pp#L83

I used a `$title` with a space in it, which is how I discovered this problem. So the shellquoting protects the exec command line, but I guess the problem is that usage of those arguments inside the [script](https://github.com/voxpupuli/puppet-selinux/blob/37e1761d6df0a02439416bcc17e384a84249465d/files/selinux_build_module_simple.sh) still needs proper quoting. What do you think?

PR #380, opened 2023-11-23
Remove legacy top-scope syntax

smortex:


kenyon:
I'm thinking we should just remove the legacy facts from the table instead. Preparing a pull request.

kenyon:
See #405 as an alternative that just removes the legacy facts column from the readme.

PR #388, opened 2024-04-12
modulesync 9.1.0

bastelfreak:
modulesync 7.4.0


.github/workflows/ci.yml
```diff
@@ -18,4 +18,4 @@ concurrency:
 jobs:
   puppet:
     name: Puppet
-    uses: voxpupuli/gha-puppet/.github/workflows/basic.yml@v1
+    uses: voxpupuli/gha-puppet/.github/workflows/beaker.yml@v2
```
ekohl, line 21: I think we should explicitly disable acceptance tests in `sync.yml`, at least for now.

ekohl (changes requested):
Can we squash the commits into a single generated commit? Essentially regenerate the whole PR.

bastelfreak:
fixed the commits

PR #391, opened 2024-06-24
use modern stdlib function

vchepkov:
It seems one invocation of ensure_packages was missed in 326d5dbaad751800f18eccd80f167835c797de6a


bastelfreak:
marking this as enhancement because we already require stdlib 9.

PR #431, opened 2026-04-30
selinux_login: Make semanage spec independent of host SELinux state

silug:
The "prefetch does not find a provider for nonexistant user" example relies on `Puppet::Type#provider` returning `nil` when prefetch did not attach a provider. That only happens when the `:semanage` provider is unsuitable on the host running the tests, because the provider is confined to `os.selinux.enabled: true` and `Puppet::Type#provider` falls back to instantiating the suitable default when no provider is set.

The Debian-based CI containers don't have SELinux, so the test passes upstream. On a host with SELinux on (Fedora, RHEL, etc.) the fallback returns a fresh `ProviderSemanage` and the assertion fails.

Stub `suitable?` to `false` at the describe level so the test reflects what prefetch actually did, regardless of host state.

N/A

### DIFF
--- a/spec/unit/provider/selinux_login/semanage_spec.rb
+++ b/spec/unit/provider/selinux_login/semanage_spec.rb
@@ -80,6 +80,15 @@ RESOURCE_EXAMPLE.delete(:source)
 LOGIN_TYPE = Puppet::Type.type(:selinux_login)
 
 describe LOGIN_TYPE.provider(:semanage) do
+  # The provider is confined to `os.selinux.enabled: true`. On a host where
+  # SELinux is enabled, the provider is "suitable" and Puppet::Type#provider
+  # falls back to instantiating it for resources that prefetch did not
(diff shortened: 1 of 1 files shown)


silug:
Without this change, on Fedora `bundle exec rake spec` fails with errors like this:
```
Failures:

  1) Puppet::Type::Selinux_login::ProviderSemanage on redhat-8-x86_64 with resources differing from the catalog prefetch does not find a provider for nonexistant user is expected to be nil
     Failure/Error: it { expect(resources['localuser2'].provider).to be_nil }
     
       expected: nil
            got: Selinux_login[localuser2](provider=semanage)
     # ./spec/unit/provider/selinux_login/semanage_spec.rb:159:in 'block (6 levels) in <top (required)>'
```

With this change, `bundle exec rake spec` works.
```
4762 examples, 0 failures
```

kenyon (approved):

